PR #7, opened 2015-01-13
fix(router): do not add routes with nav=false

tyrsius:
routes with nav:false were still being included in router.navigation

routes that contain nav:false will no longer be included


tyrsius:
`'nav' in config` ([from here](https://github.com/aurelia/router/blob/master/src%2Frouter.js#L183)) causes `nav: false` to still evaluate to true

PR #30, opened 2015-02-14
feat(router): add support for named pipelines

### LINKED ISSUE #26: Add an API to the PipelineProvider to enable the adding of pipeline steps
Add an API to add pipeline steps to the router's pipeline provider.

For example, an authentication step might be added as the very first step in the pipeline.

Alxandr:
This adds support for small named pipelines, named
route-filters, which can easily be added at given points
in the pipeline. By default, two such pipelines are
created, "authorize", and "modelbind". These acts as
"placeholders" in the pipeline, where the end user
can easily insert their own steps.

Fixes #26

### DIFF
--- a/src/app-router.js
+++ b/src/app-router.js
@@ -12,6 +12,10 @@ export class AppRouter extends Router {
     document.addEventListener('click', handleLinkClick.bind(this), true);
   }
 
+  get isRoot() {
+    return true;
+  }
+
   loadUrl(url) {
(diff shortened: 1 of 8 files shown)


EisenbergEffect:
At quick glance...this looks really nice. Would you mind sending a PR to the documentation repo that adds a section to the router docs that explains how to use this? If you can do that, I think we'll be able to merge both PRs and make a lot of people happy.

Issue #2, opened 2015-01-09
bubble rejected activation promises


jdanyow:
It would be nice if there was a way to give the router a handler for activation promise rejections.  This would reduce boilerplate error handling code in activate methods.  

Code like this:

    public activate(...): Q.Promise<any> {
        var deferred = Q.defer();

        doSomething()
            .then(() => deferred.resolve())  // yay
            .fail(reason => {
                deferred.reject(reason);     // cancel activation.
                return Q.reject(reason);     // "rethrow"
            })
            .done();  // terminate the promise.  ensure error bubbles up to the unhandled exception handler  (Q.onerror).

            return deferred.promise;
    }

Could be simplified to something like this:

    <hook error handler to router>

    public activate(...): Q.Promise<any> {
        return doSomething();
    }

tyrsius:
In chat we discussed the possibility of providing a navigation-error hook, to allow the module to subscribe to errors instead of bubbling them up.

EisenbergEffect:
@tyrsius Was this the idea of adding a new navigation callback that view models can subscribe to? Such as `activationFailed`? Or is that a different idea?

jdanyow:
If there's a way provide a hook for rejected activation lifecycle promises it would make the user code simpler.  This is the use case I was thinking of:

Often code in the activate method fails- required data is no longer on the server, the user's internet connection is down, or maybe a modal popped during activation and the user cancelled rather than continue with opening the screen.  Maybe it's a real error that should be logged, etc.  With durandal, if you wanted to handle these scenarios you needed to do two things: 

1. make sure the activate promise rejects
2. terminate the promise chain yourself with `.done` to ensure the exception bubbled to Q's unhandled rejection handler.  In this central location you'd handle well known activation issues and maybe show a message "the record no longer exists" or "cannot connect to EvilCorp servers".

I wouldn't say this was a bug in durandal, `.done` was non-standard plus how is durandal supposed to know the difference between canceled nav and a real error.

In a broader sense, in any situation where Aurelia eats a rejected promise, it would be nice if this was bubbled along with any contextual information available to allow user code to handle (or ignore) in a central location.

jdanyow:
right- since I'm unable to return a terminated promise chain from my activate method I have to do the little dance with the deferred you see in my first note in this issue.

jdanyow:
Didn't explain that well- here's what I meant:

[...]

Issue #24, opened 2015-02-04
Add Router events like in Durandal


zewa666:
I liked the emitted events from the Durandal Router, as in certain situations, especially for E2E tests, those were quite helpful to understand what the current state of page routing is.

Just for the sake of completeness I'll post them here again:
- router:navigation:complete
- router:navigation:cancelled
- router:navigation:processing
- router:route:activating
- router:route:before-config
- router:route:after-config
- router:navigation:attached
- router:navigation:composition-complete
- router:route:not-found

In case of E2E tests I'd be specially interested in the `router:navigation:complete` and `router:navigation:composition-complete` event, as they would allow to work easier with scenarios like Page transitions and dynamic composition.

tyrsius:
the `before-config` and/or `after-config` be raised if no configuration is provided, or should they be skipped?

EisenbergEffect:
Actually, let's leave those two events out of the new router for now. I think the events we probably need in the end are something like this:

* router:navigation:complete
* router:navigation:cancelled
* router:navigation:processing
* router:route:activating
* router:route:deactivating
* router:route:not-found

tyrsius:
I made the pull request, but these events did not make it in. They will require a lot more thought

router:route:activating
router:route:deactivating
router:route:not-found

zewa666:
Regarding not-found what about this place? https://github.com/aurelia/router/blob/master/src/router.js#L149

EisenbergEffect:
Yes, that could be done there.

tyrsius:
I didnt want to make modifications to the base router, so that another implementation could avoid using events. @EisenbergEffect, when I asked you agreed that changes should be kept to the app-router.

EisenbergEffect:
That's true. It's probably the only way to raise that event though. We could detect the failed promise in the app router and if it has the correct message or starts with the correct indicator, raise the event I suppose.

tyrsius:
Yea, that should work. I will give that a try later today. PWkad had another suggestion that I will add as well, but it will be after work

PR #35, opened 2015-02-22
feat(router): add nav failed event

tyrsius:
added an event for router navigation failure
Router:Navigation:Failed

should address issue #2

### DIFF
--- a/src/app-router.js
+++ b/src/app-router.js
@@ -3,13 +3,15 @@ import {History} from 'aurelia-history';
 import {Router} from './router';
 import {PipelineProvider} from './pipeline-provider';
 import {isNavigationCommand} from './navigation-commands';
+import {EventAggregator} from 'aurelia-event-aggregator';
 
 export class AppRouter extends Router {
-  static inject(){ return [Container, History, PipelineProvider]; }
-  constructor(container, history, pipelineProvider) {
+  static inject(){ return [Container, History, PipelineProvider, EventAggregator]; }
+  constructor(container, history, pipelineProvider, events) {
     super(container, history);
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
Currently the Router doesn't have a dependency on the event aggregator. I'm not sure if it should or not just yet. I'm also not sure if it should publish this globally or have its own instance of the AE that it uses. Need to think about it. Any thoughts?

tyrsius:
I don't see why the router should avoid that dependency, but I can see why it might want to use its own instance. Would you imagine that instance being shared as a property on the router?

tyrsius:
I definitely *don't* think the router should use some *other* mechanism to publish events.

EisenbergEffect:
Ok. That sounds fine. I'd like to ask you to expand this PR by implementing all the same events as Durandal's router, where possible. Do you think you could do that? Also, let's use lowercase for the event names. Sounds good?

zewa666:
[Here](https://github.com/aurelia/router/issues/24) is the issue raised for Durandal Router Events. Not sure but we should maybe think to expand it for the new Pipeline stuff. That one could use some nice events as well.

Regarding the way how Events are triggered. Custom Events would be super nice because they are easy to catch in an E2E Environment. Otherwise we maybe need to find a way how to get the EventAggregator into Protractor.

tyrsius:
I can do that, should have it this week sometime

EisenbergEffect:
@tyrsius I'm gong to close this PR for now. When you've added all the events, please go ahead and submit a new PR. Also, be sure to jspm install aurelia-event-aggregator into the router project so that it's dependencies are correctly accounted for. Thanks for working on this!

Issue #3, opened 2015-01-09
specify Viewmodel at runtime


zewa666:
It would be great if we could decide at runtime which moduleId to load, based on whatever condition is provided. The current route entry looks like this where you have to provide the moduleId as a string during configuration phase.

`````
{ route: ['','welcome'],  moduleId: 'welcome',      nav: true, title:'Welcome' },
````

Lets say though the welcome screen shows a dashboard for all users (incl. anonymous), but as soon as you login you get a different screen.

So same route path but different model implementation, as of separation of concerns. So in this case it would be helpful if the moduleId could accept a function.
As of the gitter discussion it was proposed to introduce another property, e.g. moduleStrategy, which if set would override the moduleId definition. so an example would be

````
{ route: ['','welcome'],  moduleStrategy: (options) => { 
  if(options.whatever) {
    return  "welcome";
  } else {
    return "welcome2";
  }
 }, nav: true, title:'Welcome' },
````

Perhaps even make moduleStrategy accept a Promise, which is resolved later on

EisenbergEffect:
Would you need any inputs to that function?

zewa666:
hmm maybe you're right maybe just add a object parameter (like activationData) passing all necessary information. Who knows what this could be used for.

tyrsius:
the `currentInstruction` seems reasonable

kristianmandrup:
Yes, we need this for sure :) Would be awesome to integrate this with net http-client using Fetch API :) to load modules over the wire asynchronously and thus enhance the app dynamically given some business rules/data, such as the user being logged in, upgraded to VIP member or navigating to specific part of the app etc.

EisenbergEffect:
This is implemented in master and going out in the next release.

PR #39, opened 2015-03-01
feat(vm): Specify moduleId at runtime

zewa666:
feat(vm): Specify moduleId at runtime

This adds the possiblity do define a moduleStrategy which may be a
function. If present it will be executed and will rewrite the moduleId
of the given route. See https://github.com/aurelia/router/issues/3

### DIFF
--- a/src/router-configuration.js
+++ b/src/router-configuration.js
@@ -37,6 +37,12 @@ export class RouterConfiguration{
 
   mapRoute(config) {
     this.instructions.push(router => {
+      var moduleStrategy = config.moduleStrategy;
+      if(moduleStrategy !== undefined &&
+        typeof moduleStrategy === 'function') {
+        config.moduleId = moduleStrategy(router);
+      }
+
       if (Array.isArray(config.route)) {
         var navModel = {}, i, ii, current;
 
--- a/src/router.js
+++ b/src/router.js
@@ -163,6 +163,7 @@ export class Router {
       config.viewPorts = {
         'default': {
           moduleId: config.moduleId,
+          moduleStrategy: config.moduleStrategy,
           view: config.view
         }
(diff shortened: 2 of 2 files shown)


EisenbergEffect:
I don't know if the timing of this is correct. My understanding is that people want this strategy to execute later, after the route has been matched so that it can dynamically choose the module id based on the actual navigation instruction. Can you investigate moving the strategy invocation logic?

zewa666:
Hmm I've tried at least a dozen places and all seemed too late, as the router already kicked off. The one in this PR seemed to be the only viable one. What would be the benefit of moving it next to the navigation instruction? Could you think of an example information you'd like to pick out of it? Besides that isn't currentInstruction provided by the router always an option?

EisenbergEffect:
Why not call it from here: https://github.com/aurelia/router/blob/master/src/navigation-instruction.js#L2
It's the earliest time you have the instruction obvious. Also, the moduleId wouldn't have been needed yet by the router, if I remember correctly.

zewa666:
I'm pretty sure I tried that, but will give it another shot later this afternoon and see if it works. Which params should be passed to the function?

EisenbergEffect:
Probably just need to pass `this` so that the callback has the instruction. You would need to invoke it at the bottom so that everything was initialized first. It's possible that we also need to add something to the configuration code that prevents moduleId being "inferred" during configuration when it's not present but the strategy property is.

EisenbergEffect:
Implemented a new config property `navigationStrategy`. If present, it will receive the navigation instruction which it can alter in any way, including setting the moduleId. It can also return a promise if it needs to do this asynchronously.

kristianmandrup:
Awesome!!! :)

PR #48, opened 2015-03-17
fix(Router) enable complex route config settings

jdanyow:


jdanyow:
My change doesn't assign `withChild.settings` because this is done already, a few lines below where my change is.

PR #58, opened 2015-04-05
Improved route generation support

bryanrsmith:


src/router.js
```diff
@@ -161,12 +161,19 @@ export class Router {
...
-      return this.parent.generate(name, params);
+  generate(name, params, options) {
+    options = options || {};
+    if((!this.isConfigured || !this.recognizer.hasRoute(name)) && this.parent){
+      return this.parent.generate(name, params, options);
+    }
+
+    let root = '';
+    let path = this.recognizer.generate(name, params);
+    if (options.absolute) {
```
bryanrsmith: This is probably a bad name since it's not an absolute URI. `fromRoot`?

PR #67, opened 2015-04-17
Router navigation improvements

### LINKED ISSUE #52: Make navigation to urls starting with / navigate from root
When navigating to urls that start with a `/` (like `router.navigate('/foo')`), it should navigate from the root, not the current router.

Ref #51

bryanrsmith:
* Support navigating to absolute paths from any router instance
* Add `router.navigateToRoute` convenience method, equivalent to `router.navigate(router.generate(...))`
* Minor code cleanup

fixes #52

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -6,6 +6,8 @@ import {NavigationInstruction} from './navigation-instruction';
 import {RouterConfiguration} from './router-configuration';
 import {processPotential} from './util';
 
+const isRooted = /^#?\//;
+
 export class Router {
   constructor(container, history) {
     this.container = container;
@@ -37,22 +39,7 @@ export class Router {
 
     for(var i = 0, length = nav.length; i < length; i++) {
       var current = nav[i];
-
-      if(!this.history._hasPushState){
-        if (this.baseUrl[0] == '/') {
-          current.href = '#' + this.baseUrl;
-        } else {
-          current.href = '#/' + this.baseUrl;
-        }
-      } else {
-        current.href = '/' + this.baseUrl;
-      }
-
-      if (current.href[current.href.length - 1] != '/') {
-        current.href += '/';
-      }
-
-      current.href += current.relativeHref;
+      current.href = this.createRootedPath(current.relativeHref);
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/router.js
```diff
@@ -70,16 +57,45 @@ export class Router {
...
+      path += '/';
+    }
+
+    path += this.baseUrl;
+
+    if (path[path.length - 1] != '/' && fragment[0] != '/') {
+      path += '/';
+    }
+
+    return normalizeAbsolutePath(path + fragment, this.history._hasPushState);
```
bryanrsmith: We should add a `getRoot()` method to `History` so this code can stop relying on private data.

src/router.js
```diff
@@ -70,16 +57,45 @@ export class Router {
...
+    if (isRooted.test(fragment)) {
+      fragment = normalizeAbsolutePath(fragment, this.history._hasPushState);
+    } else {
+      fragment = this.createRootedPath(fragment);
+    }
+
     return this.history.navigate(fragment, options);
   }
 
+  navigateToRoute(route, params, options) {
```
bryanrsmith: I think navigating with a route + params will be much more common than navigating with a fragment. Should we consider swapping the names, so that the shorter `navigate()` handle the more common scenario? Or maybe we can come up with a shorter name for this... `go()` is nice and short, but it would make the two methods pretty ambiguous.

bryanrsmith:
One case that still needs testing (and possible code changes) is hosting an Aurelia app from a subdirectory on the server and using pushstate. Eg, `example.com/aurelia` as the app root, with `example.com/aurelia/flickr`, etc. I don't think the current code will properly detect `/aurelia` as the root instead of `/`.

I can work on that if the rest of the changes in this PR look reasonable.

src/router.js
EisenbergEffect: Sounds good. Feel free to add that and send a PR.

src/router.js
EisenbergEffect: This is a tough call. I'm really not sure. Let's leave it as is for now.

EisenbergEffect:
Looks pretty good. If you want to work on the root issue with pushstate, please go ahead. It's been a bit since I looked at the code, but I believe there's something in history that relates to setting the root. You might want to check that out.

Also, can you comment a  bit more on the changes to createNavigationInstruction? Was there in change in behavior there or was that just cleanup work?

bryanrsmith:
Yes, the `createNavigationInstruction` changes were just to remove some work that was unnecessarily repeated. I should add tests to verify, though.

EisenbergEffect:
If you can add a few test to verify around that, I think this will be good to go.

bryanrsmith:
Done! I'll plan on working on both the `BrowserHistory._hasPushState`, and the pushstate subdirectory issues as separate PRs.

PR #76, opened 2015-04-28
fix to support viewPorts in routes without moduleId or redirect + unit test

shanonvl:
This fixes what I believe is a bug; the following code causes the `validateRouteConfig` to throw an `Error` because it's currently expecting either `moduleId` or `redirect`.  

```
this.router.configure(config => {
      config.title = 'Aurelia';
      config.map('',
        {
          nav: true,
          title: 'Welcome',
          viewPorts: {
            'default': {
              moduleId: './welcome/welcome'
            },
            'dataTables': {
              moduleId: './plugins/dataTables/sample'
            }
          }
        });
    });
```

The included unit test validates this case.

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -275,11 +275,11 @@ export class Router {
 
 function validateRouteConfig(config) {
   let isValid = typeof config === 'object'
-    && (config.moduleId || config.redirect)
+    && (config.moduleId || config.redirect || config.viewPorts)
     && config.route !== null && config.route !== undefined;
 
   if (!isValid) {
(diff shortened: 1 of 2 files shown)


shanonvl:
Thinking about this more ... I'm guessing there's no `moduleId` provided because my example refers to the top-level `App` view?   If so, I'm not sure if my fix is ideal vs. defining a `moduleId` for the `App` view ?

EisenbergEffect:
I think your fix works. @bryanrsmith Would you review this? I'm going to give you write access so you can adjust issues via the zenhub ui.

bryanrsmith:
Looks good to me!

PR #78, opened 2015-04-28
Move activation strategy constants into "enum" object

bryanrsmith:


src/navigation-plan.js
```diff
@@ -1,8 +1,10 @@
 import {Redirect} from './navigation-commands';
 
-export var NO_CHANGE = 'no-change';
-export var INVOKE_LIFECYCLE = 'invoke-lifecycle';
-export var REPLACE = 'replace';
+export const activationStrategy = {
```
bryanrsmith: This is named like an enum type, but maybe the plural `activationStrategies` would be more appropriate.

EisenbergEffect: I think we should keep it singluar. I would rather update the logger to match that.

PR #81, opened 2015-05-04
Refactor navModel

bryanrsmith:
This change refactors the creation and configuration of navModels during route registration. `navModel.setTitle()` can now be used to update the document's title at any time from an active navModel.


src/router.js
```diff
@@ -208,19 +217,9 @@ export class Router {
 
     config.navModel = navModel;
 
-    if ((config.nav || 'order' in navModel) && this.navigation.indexOf(navModel) === -1) {
-      navModel.order = navModel.order || config.nav;
-      navModel.href = navModel.href || config.href;
-      navModel.isActive = false;
-      navModel.config = config;
```
bryanrsmith: This was unused, and is no longer copied onto the navModel. The same is true for `settings` above.

EisenbergEffect: Is that just due to a refactoring? I want to make sure that all of this is still available as part of the new nav model. Devs need it to bind to in their UI. Also, settings needs to be present since that's considered a "safe place" for developer to put a custom object or data.

bryanrsmith: Makes sense. I will restore them and add comments clarifying their use.

bryanrsmith: Restored.

EisenbergEffect:
@bryanrsmith Would you mind rebasing your changes on the latest master? Things are a little bit messy in this PR ;)

bryanrsmith:
Oops. Certainly :-)

EisenbergEffect:
@bryanrsmith I'll close this PR and let you open a new one after that's all fixed up.

bryanrsmith:
`for .. of`'s removed. Looks like TS added support in 1.5, but Babel does generate a lot of code for it here. Looks like Babel is smart & transforms it to a simple `for .. in` if the collection is an array literal, but that's rarely what we have :-)

EisenbergEffect:
@bryanrsmith Did you intend for this to be in the release today? or wait on this until the next release?

bryanrsmith:
This one should wait for a larger release.

EisenbergEffect:
Ok, I thought so, but just wanted to make sure that was your thinking as well.

PR #87, opened 2015-05-06
fix(pipeline): fix bug preventing errors in pipeline steps from being handled

### LINKED ISSUE #85: Ensure errors in pipeline steps are handled
Reported via Gitter at https://gitter.im/Aurelia/Discuss?at=5547f07e26b588274527ad46

We should ensure that errors in pipeline steps are well-handled.

bryanrsmith:
fixes #85

AppRouter's [pipeline error handling](https://github.com/aurelia/router/blob/master/src/app-router.js#L72-L75) is designed to handle Promises that *resolve* with error results (not rejections). When `next.reject` returns a Promise *rejection* the error is intercepted by [catch handlers](https://github.com/aurelia/router/blob/master/src/navigation-plan.js#L83) in a couple places, which wrap the result in a second cancelled result and resolve. The `result` in AppRouter's continuation ends up being a `cancelled` result, whose `output` is an inner `rejected` result, which is handled as a nav cancellation instead of an error.

An alternative solution might be to keep the `Promise.reject()`, remove some or all of the `.catch()` handlers, and move `AppRouter`'s pipeline error handling out of the `.then()` into a `.catch()`. I'm not yet confident that it would be safe to remove the necessary `.catch()`s, and would need more time to study the code.

### DIFF
--- a/src/pipeline.js
+++ b/src/pipeline.js
@@ -78,7 +78,7 @@ export class Pipeline {
     next.reject = error => {
       next.status = pipelineStatus.rejected;
       next.output = error;
-      return Promise.reject(createResult(ctx, next));
+      return Promise.resolve(createResult(ctx, next));
     };
 
     next.status = pipelineStatus.running;


EisenbergEffect:
Haven't looked at the code just yet, but I wanted to make an important note on any changes here. Whatever happens in terms of errors, the router cannot be left in an invalid state. If it is, then the entire application can break. So, we want to report errors but make sure the other navigation attempts subsequent to the error continue to work.

davidpeden3:
Ran into this today. Looking forward to the fix getting merged!

PR #91, opened 2015-05-08
fix(router): using navigationStrategy throws a validation error

bfil:
Trying to add a navigationStrategy to a route throws the following validation error:

`You must have at least a route and a moduleId, redirect, or viewPorts. at validateRouteConfig`

I've added the condition to the validation and made other changes to make the following route configuration work:

```
{
    route: ['', 'welcome'], nav: true, title: 'Welcome',
    navigationStrategy: function(instruction) {
        instruction.config.moduleId = 'welcome';
    }
}
```

The changes I've made fix some of the issues I encountered, but I'm not 100% sure it's the way that this is supposed to work, please review.

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -146,8 +146,8 @@ export class Router {
 
       if (typeof first.handler === 'function') {
         return evaluateNavigationStrategy(instruction, first.handler, first);
-      } else if(first.config && 'navigationStrategy' in first.config){
-        return evaluateNavigationStrategy(instruction, first.config.navigationStrategy, first.config);
+      } else if(first.handler && 'navigationStrategy' in first.handler){
+        return evaluateNavigationStrategy(instruction, first.handler.navigationStrategy, first.handler);
       }
 
       return Promise.resolve(instruction);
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
You did well there :) Thank you!

bryanrsmith:
Looks good to me :-)

You say this fixes "some of the issues" you encountered. What are the remaining issues you're seeing?

bfil:
It seems like it's working fine for me after this change.

It's just that the fact that it was checking for `first.config` instead of the `first.handler` made me think that maybe I was doing something wrong, but if you are happy with it, then it's alright, it's working fine for me with this fix.

EisenbergEffect:
I've merged this. I think it should solve the immediate issue. @bfil If you have other issues, please open up tickets for those. Thanks for this fix. This is what I get for trying to implement a "quick" feature a week or so ago...

bfil:
Yeah I understand that, I would have liked to add tests for that, but no time now, maybe I'll do it later on.

Issue #90, opened 2015-05-08
Support eagerly configuring child routers


bryanrsmith:
Lazily configuring child routers has a couple downsides:

1. We can't support hierarchical nav menus because the application's route structure isn't known until all child routes are activated.
2. We can't support generating URLs for routes in "sibling" routers for the same reason.

If we let child routers be eagerly configured we could implement features that require knowledge of the whole apps routing structure. Ideas:

1. We could let route configs specify child router configs inline, so an App module could configure a larger hierarchy of routes.
2. We could let route config objects indicate that their modules will configure a child router that should be eager-loaded.

sorenhoyer:
+1 this would indeed be a nice enhancement. :)

aozisik:
I would love this, as well. Until then, is there a workaround one could use?

npelletm:
+1 for this feature :)

larsfjerm:
+1 - fingers crossed for this!

Cedware:
+1 configuring routes in the ViewModel isn't really pretty

dlo83:
+1

npelletm:
For the next RC could we expected this feature ?

claylaut:
+1

bryanrsmith:
Sorry, but I don't know of any plans to work on this before 1.0. I really want it too, but I haven't had much time to spare lately. If anyone wants to work on a PR I'd be more than happy to discuss the necessary changes and answer questions.

ngelx:
+1

bhatnagar-ankur:
+1

sysanin:
+1, any plans on this feature?

EisenbergEffect:
We are working towards v1 release in the next few weeks. After that we'll look at adding some new features to the router. This is high on the list.

(As this is open source, all community members are welcome to implement this and send a PR. That helps to move certain requests ahead faster.)

heruan:
I think the solution for this and https://github.com/aurelia/router/issues/89 should address a more general scenario. Aurelia is so modular it comes natural to have an app built upon different external modules which may contribute to the route hierarchy. For example:
```
/       (main-app)
/admin  (admin-app)
/blog   (blog-app)
/wiki   (wiki-app)
```
[...]

Issue #89, opened 2015-05-08
Router.generate should support routes registered with sibling routers


bryanrsmith:
Currently `router.generate()` will search for named routes registered with the current router, or an ancestor router. Consider a StackOverflow-style site with a `Questions` VM that configures a child router with 'list', 'search', and 'details' routes, and a `Users` VM that configures a child router with 'list', 'search', and 'profile' routes. The question details page will need to link to the user profile page of the asker, and the user profile page will need to link to the question details for questions asked and answered by the user. Currently this can only be done by manually constructing an absolute URL using string concatenation. It would be great if `router.generate()` could handle routes registered in other branches of the router tree.

bryanrsmith:
What if we supported route namespaces, so that from the `Questions` VM, I could `router.generate('users.details', { id: '123' })`. This would activate the 'users' module if it hasn't been configured yet, and then look for the 'details' route under that child router.

bryanrsmith:
After #90 we should be able to support generating for eagerly configured routers.

larsfjerm:
I've been waiting for a fix for this in a long time now. Would this be possible when the beta comes out? I would say it's a crucial functionality.

bryanrsmith:
I agree. I really want this too. I don't think it will be ready before
beta, though. It requires a pretty big internal change that I haven't had
time for. Hoping to get to it soon after beta.

larsfjerm:
Nice to hear. You've all done a great job so far. Looking forward for the rest!

migajek:
Any news on this?
I think this is a must-have for the complex routing solution.
I'm from Angular background and I really like their's de-facto standard routing solution, ui router.

It's hierarchy is defined through namespacing, following/extending a bit @bryanrsmith 's example:

```
Questions 
   Questions.list
   Questions.search
   Questions.details
Users
   Users.list
   Users.search
   Users.details
      Users.details.bio
      Users.details.experience
```

the feature I love there is relational navigation, very intuitive.

Assuming ``Users.details`` is our active (current) route:
``^.list`` == ``Users.list``
``.experience`` == ``Users.details.experience``

I believe the example is self-explanatory ;)

This hierarchical namespacing approach has several major upsides; 
it doesn't require the user to type the whole "path" each time, just the relative part - both when configuring and referencing the route
it makes it easier to maintain complex apps (CRUD especially, where one have tens of "list", "detail", "create", "update" routes)

I'd love to see something like that in Aurelia.

peisenmann:
[...]

Issue #50, opened 2015-03-18
Router refactoring


Alxandr:
I named this issue rather ambiguously because it will be covering a few points that I find somewhat hard to take apart, or name properly.

Currently, when you call `router.navigate` it does all of the work, and builds a instruction that is sent through the router pipeline. The `Redirect` navigation command is a simple wrapper around this call.

I propose inverting this relationship, and renaming `Redirect` to `NavigateCommand`, and adding a new method to the router named `process`, that would work as following:

The `router.navigate` would be turned into a simple convenience method, simply calling

```javascript
navigate(url, opts) {
  var cmd = new NavigateCommand(url, opts, this);
  this.process(cmd);
}
```

The `process` command would simply forward the call to the App Router, which in turn would invoke

```javascript
process(navigationCommand) {
  navigationCommand.navigate(this);
}
```

And the functionality that is currently in the `navigate` method would be moved to the `NavigateCommand` class instead. This would enable one to create their own navigation commands, that did other things (and could even extend `NavigateCommand` to make things such as `NavigateToLoginCommand` that handled capturing return address etc.).

Alxandr:
Oh, and btw, `RedirectCommand` should also be added (given that it does redirect, instead of navigate). However, with this architecture it could be created really simple like this:

```javascript
class RedirectCommand extends NavigateCommand {
  constructor(url) {
    super(url, { trigger: true, replace: true });
  }
}
```

Alxandr:
Oh, and a point that just popped up from a user that was trying to call `navigate` from within `activate`, any call to `process` should cancel out any currently ongoing navigation commands.

EisenbergEffect:
@bryanrsmith After the current set of changes you are working on, I'd like to look again at this issue and see if this would be advantageous and what the level of effort would be.

EisenbergEffect:
@bryanrsmith Please make a final review of this and see if we want to do this re-design or if you think it's ok the way things are.

bryanrsmith:
I do like the proposal, and we could do it fairly easily, but I think it's relatively low value at this point. router.navigate doesn't really have much control either-- it delegates directly to the history implementation. I think we can leave it the way it is for now. We can make it a nonbreaking change if we decide to do it in the future.

PR #96, opened 2015-05-10
Replace route config property derivations with stricter validation

bryanrsmith:
BREAKING CHANGE: This is a breaking change to anyone relying on the `RouterConfiguration` to derive values for required route config properties. All config properties must now be specified explicitly.

### DIFF
--- a/src/router-configuration.js
+++ b/src/router-configuration.js
@@ -25,7 +25,7 @@ export class RouterConfiguration{
       let routeConfigs = [];
 
       if (Array.isArray(config.route)) {
-        for (let i = 0, ii = config.route.length; i < ii; i++) {
+        for (let i = 0, ii = config.route.length; i < ii; ++i) {
           let current = Object.assign({}, config);
           current.route = config.route[i];
           routeConfigs.push(current);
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
@bryanrsmith You are doing great work on the router. Keep it up! I'll get this merged in a few minutes. Any interest in writing something for the official blog about the router? Maybe to be published in a few weeks after we finish all this work up? Also, any more breaking changes planned that will affect the public api? K Scott is prepping Pluralsight courses, so we want to make sure he doesn't have to change his videos.

bryanrsmith:
@EisenbergEffect Thanks! I can definitely do a blog post. I don't have any other plans for breaking changes, but some of the breakages have come after I've discovered problems or dug into new feature areas. I'd love to get more tests in place so we can be more confident about stability of the current feature set. #50, #89, and #90 could possibly result in significant API or feature changes (not necessarily breaking), but need further planning work.

I don't really get any extended coding sessions because of the baby, so I've mostly been hitting the smaller issues. Will try to start tackling some of the important larger issues soon.

EisenbergEffect:
Sounds good. Keep me posted. If we can keep changes to being either "addative" or if it has to be a break, hopefully in some area that isn't widely used yet. Mainly, I imagine Scott will use the router to configure routes and do route generation, so we should make sure that is stable asap so he doens't have to re-record his videos. The more advanced stuff probably won't be touched on. I'm going to try to stay up with what he's doing so we can know what api's he's about to record on and make sure we are good on those before he does it.

PR #97, opened 2015-05-12
fix(router): cancelling a navigation instruction with a redirect raises the wrong navigation event

bfil:
Using the following `AuthorizeStep` (provided as an example in the Aurelia documentation):

```javascript
class AuthorizeStep {
  run(routingContext, next) {
    if (routingContext.nextInstructions.some(i => i.config.auth)) {
      var isLoggedIn = /* insert magic here */false;
      if (!isLoggedIn) {
        return next.cancel(new Redirect('login'));
      }
    }
    return next();
  }
}
```

When cancelling the navigation instruction with a redirect, a `router:navigation:complete` event is raised for the cancelled instruction, instead of the expected `router:navigation:cancelled`.

### DIFF
--- a/src/app-router.js
+++ b/src/app-router.js
@@ -58,7 +58,7 @@ export class AppRouter extends Router {
     var context = this.createNavigationContext(instruction);
     var pipeline = this.pipelineProvider.createPipeline(context);
 
-    pipeline.run(context).then(result => {
+    return pipeline.run(context).then(result => {
       this.isNavigating = false;
 
       if (!(result && 'completed' in result && 'output' in result)) {
@@ -78,13 +78,20 @@ export class AppRouter extends Router {
         result.output.navigate(this);
       } else if (!result.completed) {
         this.navigate(this.history.previousFragment || '', false);
-        this.events.publish('router:navigation:cancelled', instruction)
+        this.events.publish('router:navigation:cancelled', instruction);
       }
 
       instruction.resolve(result);
       this.dequeueInstruction();
+      return result;
     })
-    .then(result => this.events.publish('router:navigation:complete', instruction))
+    .then(result => {
+      if(result.completed) {
+        this.events.publish('router:navigation:complete', instruction);
+      } else {
+        this.events.publish('router:navigation:cancelled', instruction);
+      }
+    })
     .catch(error => {
       console.error(error);
     });


bryanrsmith:
I think 'router:navigation:complete' is intended to indicate that the router's navigation processing has finished, and not necessarily that the navigation was successful. So for a cancelled navigation you'd expect to see the 'cancelled' event followed by the 'complete' event. Are you seeing both events, or only the 'complete'?

I'm not 100% certain of the intended behavior, but I suspect the change we'll want is to leave the code as it is, but add in a 'success' event. That way you could easily choose to listen for any possible navigation result. /cc @EisenbergEffect

EisenbergEffect:
I think we definitely need to make this clearer. "complete" means that the process has finished, but I think we probably do need success/fail events so we can make those cases clearer.

bfil:
Yeah probably a success/fail events would be enough to wire in some listeners that should trigger only when a navigation completed successfully.

The code as it stands when a navigation instruction is cancelled it does create a result object containing:

```
{
    completed: false,
    status: "cancelled"
}
```

For this reason I thought that the cancelled event should be raised instead.

And to answer @bryanrsmith it is NOT raising the cancelled event at all, only the complete, because of the following piece of code:

```
if (isNavigationCommand(result.output)) {
    result.output.navigate(this);
} else if (!result.completed) {
    this.navigate(this.history.previousFragment || '', false);
    this.events.publish('router:navigation:cancelled', instruction);
}
```

It always enters the `if`, because the `Redirect` is a `NavigationCommand`.

EisenbergEffect:
@bryanrsmith I'm gong to assign this to you and let you review more thoroughly. Let me know if this should be merged or if there's a more nuanced approach you need to take to get the correct behavior.

EisenbergEffect:
@bryanrsmith Just wanted to check in with you and see where we are on this. Thanks for all the great work you're doing on the router. I got your other PRs in just now.

bryanrsmith:
@EisenbergEffect Think I'm almost done with a fix for #103 & then I'll take a look. I should be able to get to it today.

bryanrsmith:
So, I can add a 'success' event pretty easily. It would basically just be published when 'cancelled' or 'error' were not published. I'm not sure that's good enough, though. 

Currently navigations can, and often do, consist of multiple navigations under the hood.

* If you return a redirect from an `activate`, a new navigation is started before the first is completed. So you get: processing(a), processing(b), complete(a), complete(b). Note that in this case we don't publish a cancelled event.
* If you cancel a navigation by returning a rejected promise from `activate` (with no redirect), the router tries to revert to the previous state. This also starts a new navigation, so you get: processing(b), cancelled(b), complete(b), processing(a), complete(a).

Will simply adding a 'success' event be useful? It seems to me that it's pretty hard to tell what actually happened from this event stream alone. We might want to rethink how these events are triggered entirely. Should we simplify it by collapsing all of the "inner" navigations, so that you always see exactly 3 events per attempted navigation, regardless of how many redirects or cancellations actually happened? The two cases above might look more like..

* processing(a), success(b), complete(b)
* processing(b), cancelled(b), complete(b)

This seems more intuitive and more useful to me, but it'll probably take a bit of work because it doesn't really reflect the way the router works internally. We could record the inner navigations and attach the list to one or more of the events, so consumers can see how the sausage was made if they really want to.

What do you guys think?

EisenbergEffect:
@bryanrsmith Yes. I think you are describing the behavior that we actually want.

bfil:
@bryanrsmith I agree that would make more sense, and would make it easier to handle the two cases (success/cancelled)

bryanrsmith:
I've opened #130 to address this issue with a more thorough refactor. I'll go ahead and close this one. Thanks!

kristianmandrup:
@bryanrsmith The solution you propose looks elegant and much appreciated :) Thanks!

PR #101, opened 2015-05-17
fix(router): handle navigating to the empty string in a child router

### LINKED ISSUE #99: navigating to default route in child router navigates to default route in parent
possible router bug?  i have a child router who's first route is "".  the child router belongs to a non-default route in the app router.  when i try to navigate the child router to its default route, the app navigates to the default route in the app router

```javascript
this.router.navigate("", true);
```

bryanrsmith:
fixes #99

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -84,10 +84,6 @@ export class Router {
       return this.parent.navigate(fragment, options);
     }
 
-    if (fragment === '') {
-      fragment = '/';
-    }
-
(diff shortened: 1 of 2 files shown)


src/router.js
```diff
@@ -84,10 +84,6 @@ export class Router {
       return this.parent.navigate(fragment, options);
     }
 
-    if (fragment === '') {
```
bryanrsmith: This case is now handled by `createRootedPath`.

PR #109, opened 2015-05-25
fix(navigation-plan): handle redirects in child router configs

### LINKED ISSUE #103: Redirect configs in child routers don't work
The router does not write the "Welcome" into the URL.

The angularjs router has the same "design flaw".

Can this be fixed?

My expectation is to get a correct URL for default routes. Also for the initial root route.

image

bryanrsmith:
This change fixes several issues with configs specifying redirects.
 * The pipeline now properly handles route configs specifying redirects in child routers. Previously this was only handled for the top-level routes in `BuildNavigationPlanStep`.
* A redirect location of '' is now accepted.
* Redirect locations are now resolved relative to the navigation instruction rather than assumed to be relative.
* Query strings are now preserved across redirects.

fixes #103

### DIFF
--- a/src/navigation-plan.js
+++ b/src/navigation-plan.js
@@ -1,4 +1,5 @@
 import {Redirect} from './navigation-commands';
+import {resolveUrl} from './util';
 
 export const activationStrategy = {
   noChange: 'no-change',
@@ -10,6 +11,15 @@ export function buildNavigationPlan(navigationContext, forceLifecycleMinimum) {
   var prev = navigationContext.prevInstruction;
   var next = navigationContext.nextInstruction;
   var plan = {}, viewPortName;
+
+  if ('redirect' in next.config) {
+    let redirectLocation = resolveUrl(next.config.redirect, getInstructionBaseUrl(next));
+    if (next.queryString) {
+      redirectLocation += '?' + next.queryString;
+    }
+
+    return Promise.reject(new Redirect(redirectLocation));
+  }
 
   if (prev) {
     var newParams = hasDifferentParameterValues(prev, next);
@@ -71,17 +81,13 @@ export function buildNavigationPlan(navigationContext, forceLifecycleMinimum) {
 }
 
 export class BuildNavigationPlanStep {
-	run(navigationContext, next) {
-    if (navigationContext.nextInstruction.config.redirect) {
-      return next.cancel(new Redirect(navigationContext.nextInstruction.config.redirect));
-    }
-
+  run(navigationContext, next) {
     return buildNavigationPlan(navigationContext)
       .then(plan => {
         navigationContext.plan = plan;
         return next();
       }).catch(next.cancel);
-	}
+  }
 }
 
 function hasDifferentParameterValues(prev, next) {
@@ -100,4 +106,14 @@ function hasDifferentParameterValues(prev, next) {
   }
 
   return false;
+}
+
+function getInstructionBaseUrl(instruction) {
+    let instructionBaseUrlParts = [];
+    while(instruction = instruction.parentInstruction) {
+      instructionBaseUrlParts.unshift(instruction.getBaseUrl());
+    }
+
+    instructionBaseUrlParts.unshift('/');
+    return instructionBaseUrlParts.join('');
 }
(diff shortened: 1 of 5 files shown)


src/navigation-plan.js
```diff
@@ -11,6 +12,15 @@ export function buildNavigationPlan(navigationContext, forceLifecycleMinimum) {
   var next = navigationContext.nextInstruction;
   var plan = {}, viewPortName;
 
+  if ('redirect' in next.config) {
+    let redirectLocation = resolveUrl(next.config.redirect, getInstructionBaseUrl(next));
```
bryanrsmith: We need to resolve the `next.config.redirect` relative to the router it was registered in, but the router's `baseUrl` isn't set until a navigation is committed in it. Instead we walk up the chain of current instructions and build up the baseUrl to use for the redirect.

src/navigation-plan.js
```diff
@@ -11,6 +12,15 @@ export function buildNavigationPlan(navigationContext, forceLifecycleMinimum) {
...
+    if (next.queryString) {
+      redirectLocation += '?' + next.queryString;
+    }
+
+    return Promise.reject(new Redirect(redirectLocation));
```
bryanrsmith: We need to do this in `buildNavigationPlan` because a redirect could be configured in any node of the tree. The redirect must be triggered as soon as we find one to avoid trying to build a plan for a navigation that won't happen (a module ID is needed for that, which we don't have).

src/navigation-plan.js
```diff
@@ -71,17 +81,13 @@ export function buildNavigationPlan(navigationContext, forceLifecycleMinimum) {
...
-    if (navigationContext.nextInstruction.config.redirect) {
-      return next.cancel(new Redirect(navigationContext.nextInstruction.config.redirect));
-    }
-
+  run(navigationContext, next) {
     return buildNavigationPlan(navigationContext)
       .then(plan => {
         navigationContext.plan = plan;
         return next();
       }).catch(next.cancel);
```
bryanrsmith: This `catch(next.cancel)` will now trigger the redirect previously handled above.

src/navigation-plan.js
```diff
@@ -101,3 +107,13 @@ function hasDifferentParameterValues(prev, next) {
 
   return false;
 }
+
+function getInstructionBaseUrl(instruction) {
```
bryanrsmith: The baseUrl we need is the concatenation of all ancestor instruction's baseUrls. Is this guaranteed to be correct?

src/util.js
```diff
@@ -16,4 +16,43 @@ export function processPotential(obj, resolve, reject){
...
+export function resolveUrl(fragment, baseUrl, hasPushState) {
+  if (isRootedPath.test(fragment)) {
+    return normalizeAbsolutePath(fragment, hasPushState);
+  } else {
+    return createRootedPath(fragment, baseUrl, hasPushState);
+  }
+}
+
+const isRootedPath = /^#?\//;
+const isAbsoluteUrl = /^([a-z][a-z0-9+\-.]*:)?\/\//i;
```
bryanrsmith: This stuff just moved out of `router` for reuse in `navigation-plan`. The only changes were to parameterize the baseUrl and hasPushState instead of reading them off the current instance.

bryanrsmith:
@EisenbergEffect I'd appreciate a thorough review on this one. Does this seem like a good way to solve the problem, and are there are edge cases I'm not accounting for? (We should write more tests...)

EisenbergEffect:
@bryanrsmith I looked over the code. It looks pretty good to me. How do you feel about it? Ready for me to merge it in? Did you check it with the skeleton and contacts demos?

EisenbergEffect:
@bryanrsmith Just want to double check with you. It looks good to me. Let me know.

bryanrsmith:
Sorry, I didn't have time to double check it last night. I did test with
skeleton nav, but not contacts. I think it's good, but I can double check
app-contacts tonight if you'd like to wait.

EisenbergEffect:
It can wait until the evening. No problem at all.

bryanrsmith:
@EisenbergEffect Ok, I think this is ready.

PR #115, opened 2015-05-29
fix(router): incorrectly comparing routes parameter values in navigation plan

bfil:
When checking if the routes parameters changed when navigating between views, the logic does not take into account when the next route parameters does not contain parameters from the previous one.

It works fine in this example: 
Current view params: `{ param1: 'someValue' }`
Next view params: `{ param1: 'someValue', param2: 'anotherValue' }`

It does not work in the opposite direction:
Current view params: `{ param1: 'someValue', param2: 'anotherValue' }`
Next view params: `{ param1: 'someValue' }`

because it iterates through only the next route parameters, comparing `param1` without noticing that `param2` is now missing compared to the previous view.

There might be a smarter way to compare the two objects, so it might be better refactored, but wasn't sure of how the wildcards should be handled.

### DIFF
--- a/src/navigation-plan.js
+++ b/src/navigation-plan.js
@@ -99,5 +99,11 @@ function hasDifferentParameterValues(prev, next) {
     }
   }
 
+  for (var key in prevParams) {
+    if (prevParams[key] !== nextParams[key]) {
+      return true;
+    }
+  }
+
   return false;
 }
--- a/src/navigation-plan.js
+++ b/src/navigation-plan.js
@@ -90,16 +90,20 @@ function hasDifferentParameterValues(prev, next) {
       nextWildCardName = next.config.hasChildRouter ? next.getWildCardName() : null;
 
   for (var key in nextParams) {
-    if (key == nextWildCardName) {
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
@bryanrsmith Can you review this?

bryanrsmith:
Thanks, @bfil. Good catch.

I think we need to apply the same wildcard name check to prev that's being used on next, since the previous instruction could also have a child router. It would also be nice to use the same comparison operator in both cases-- your new code uses strict equality, but the old code did not. (strict equality does seem more appropriate to me here). Would you mind making those changes? I think this will be good to go after that.

bfil:
I've added the changes you requested, I was not sure about the wildcard one, I guess it makes sense to have it then. I still don't like how the code repeats itself here, but it's alright. :)

EisenbergEffect:
Ok, this looks good to me. @bryanrsmith What do you think? Ready to go?

bryanrsmith:
You could factor out a little helper function if the duplication bothers you, but I think it's fine. :+1:

PR #130, opened 2015-06-21
Refactor dequeueInstruction to improve nav events and error handling

bryanrsmith:
* All navigations now raise 3 events: `processing`, (`success` | `error` | `canceled`), and `complete`.
* All navigation events include the same event args: `{ instruction, result }`, except for `processing`, which omits the result property.
* A maximum of 10 instruction executions are attempted before giving up with a navigation error.
* Unexpected return values from the pipeline are now handled as error results instead of crashing.

### DIFF
--- a/src/app-router.js
+++ b/src/app-router.js
@@ -58,8 +58,12 @@ export class AppRouter extends Router {
 
       if (!instructionCount) {
         this.events.publish('router:navigation:processing', { instruction });
+      } else if (instructionCount === this.maxInstructionCount - 1) {
+        logger.error(`${instructionCount + 1} navigation instructions have been attempted without success. Restoring last known good location.`);
+        restorePreviousLocation(this);
+        return this.dequeueInstruction(instructionCount + 1);
       } else if (instructionCount > this.maxInstructionCount) {
-        throw new Error(`Maximum navigation attempts exceeded. ${this.maxInstructionCount} navigation instructions have been attempted without success. Giving up.`);
+        throw new Error(`Maximum navigation attempts exceeded. Giving up.`);
       }
 
       let context = this.createNavigationContext(instruction);
--- a/src/app-router.js
+++ b/src/app-router.js
@@ -204,7 +204,7 @@ function resolveInstruction(instruction, result, isInnerInstruction, router) {
 function restorePreviousLocation(router) {
   let previousLocation = router.history.previousLocation;
   if (previousLocation) {
-    router.navigate(router.history.previousLocation, false);
+    router.navigate(router.history.previousLocation, { trigger: false, replace: true });
   } else {
     console.error('Router navigation failed, and no previous location could be restored.');
   }
--- a/src/app-router.js
+++ b/src/app-router.js
@@ -1,4 +1,5 @@
 import core from 'core-js';
+import * as LogManager from 'aurelia-logging';
 import {Container} from 'aurelia-dependency-injection';
 import {History} from 'aurelia-history';
 import {Router} from './router';
@@ -6,6 +7,8 @@ import {PipelineProvider} from './pipeline-provider';
 import {isNavigationCommand} from './navigation-commands';
 import {EventAggregator} from 'aurelia-event-aggregator';
 import {RouterConfiguration} from './router-configuration';
+
+const logger = LogManager.getLogger('app-router');
 
 export class AppRouter extends Router {
   static inject(){ return [Container, History, PipelineProvider, EventAggregator]; }
@@ -25,7 +28,7 @@ export class AppRouter extends Router {
     return this.createNavigationInstruction(url)
       .then(instruction => this.queueInstruction(instruction))
       .catch(error => {
-        console.error(error);
+        logger.error(error);
         restorePreviousLocation(this);
       });
   }
@@ -206,6 +209,6 @@ function restorePreviousLocation(router) {
   if (previousLocation) {
     router.navigate(router.history.previousLocation, { trigger: false, replace: true });
   } else {
-    console.error('Router navigation failed, and no previous location could be restored.');
+    logger.error('Router navigation failed, and no previous location could be restored.');
   }
 }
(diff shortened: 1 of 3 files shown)


src/app-router.js
```diff
@@ -22,14 +23,7 @@ export class AppRouter extends Router {
 
   loadUrl(url) {
     return this.createNavigationInstruction(url)
-      .then(instruction => this.queueInstruction(instruction))
-      .catch(error => {
```
bryanrsmith: This error handling is now rolled into `dequeueInstruction`.

I think I need to add something back here if `createNavigationInstruction` rejects, such as when a route is not found. We should restore the previous fragment in that case to make sure the document location remains consistent.

src/app-router.js
```diff
@@ -40,53 +34,37 @@ export class AppRouter extends Router {
...
-    }
-
-    this.isNavigating = true;
-    this.events.publish('router:navigation:processing', instruction);
-
-    var context = this.createNavigationContext(instruction);
-    var pipeline = this.pipelineProvider.createPipeline(context);
+  dequeueInstruction(instructionCount = 0) {
+    return Promise.resolve().then(() => {
+      if (this.isNavigating && !instructionCount) {
```
bryanrsmith: "inner" instructions, such as redirects and instructions added to the queue during other navigations are now considered part of the "outer" navigation, instead of completing the current instruction before moving on to the next.

src/app-router.js
```diff
@@ -40,53 +34,37 @@ export class AppRouter extends Router {
...
-        this.history.previousFragment = instruction.fragment;
-      }
+      this.isNavigating = true;
 
-      if (result.output instanceof Error) {
-        console.error(result.output);
-        this.events.publish('router:navigation:error', { instruction, result });
+      if (!instructionCount) {
+        this.events.publish('router:navigation:processing', { instruction });
+      } else if (instructionCount > this.maxInstructionCount) {
```
bryanrsmith: We could try to restore the last known good fragment instead of just giving up, but this helps avoid the more immediately problem of getting stuck in an infinite redirect loop.

src/app-router.js
```diff
@@ -172,3 +150,53 @@ function targetIsThisWindow(target) {
     targetWindow === '_self' ||
     (targetWindow === 'top' && window === window.top);
 }
+
+function processResult(instruction, result, instructionCount, router) {
+  if (!(result && 'completed' in result && 'output' in result)) {
+    resut = result || {};
+    result.output = new Error(`Expected router pipeline to return a navigation result, but got [${JSON.stringify(result)}] instead.`);
```
bryanrsmith: This threw an error that wasn't handled before. Now it's rolled up like any other navigation error.

src/app-router.js
```diff
@@ -172,3 +150,53 @@ function targetIsThisWindow(target) {
...
+    result.output.navigate(router);
+  } else {
+    finalResult = result;
+
+    if (!result.completed) {
+      let previousFragment = router.history.previousFragment;
+      if (previousFragment) {
+        router.navigate(router.history.previousFragment || '', false);
+      } else {
+        console.error('Router navigation failed, and no previous location could be restored.', result.output);
```
bryanrsmith: We should use aurelia/logging instead of writing directly to the console.

src/app-router.js
```diff
@@ -172,3 +150,53 @@ function targetIsThisWindow(target) {
...
+
+  if (!isInnerInstruction) {
+    router.isNavigating = false;
+    let eventArgs = { instruction, result };
+    let eventName;
+
+    if (result.output instanceof Error) {
+      eventName = 'error';
+    } else if (!result.completed) {
+      eventName = 'canceled';
```
bryanrsmith: I changed the inflection here to the American English version because it seems more common in other frameworks and libraries.

src/app-router.js
```diff
@@ -172,3 +150,53 @@ function targetIsThisWindow(target) {
...
+    } else {
+      router.history.previousFragment = instruction.fragment;
```
bryanrsmith: We probably need to include the query string here.

bryanrsmith:
Not quite ready, but I'll leave this open for review.

src/app-router.js
EisenbergEffect: Sounds good. If you can make that one fix, I think it's good to go. I like the direction you are going with all this.

src/app-router.js
EisenbergEffect: Sounds good. That was a bit wonky before. I like the consistency you are adding to the design. Keep it up!

src/app-router.js
EisenbergEffect: I think we will want to try to restore the last fragment. Besides avoiding an infinite loop, we want to make sure we don't leave the app in an invalid state.

src/app-router.js
EisenbergEffect: Cool!

src/app-router.js
EisenbergEffect: Yes. Please add the logging dependency and begin using that. Thanks! (I should have done that a long time ago...)

src/app-router.js
EisenbergEffect: Sounds good.

src/app-router.js
EisenbergEffect: Oops! Yes.

EisenbergEffect:
This work looks great @bryanrsmith I really appreciate how you've taken my original implementation and really cleaned it up. Keep it up. Let me know when you think this PR is ready to merge.

bryanrsmith:
@EisenbergEffect Thanks! Glad you like it. I've added several new commits to address the remaining issues I could find. I think it's ready, but if you could give it another skim that'd be great.

EisenbergEffect:
Ok, looks good to me. Things are starting to look pretty solid. Are all the tests passing? Also, did you get a chance to run this against the nav skeleton and the contacts app?

bryanrsmith:
Yep-- tests pass, and skeleton and contacts apps look good. I haven't added any new tests for it, but I'm planning to write up more tests next.

EisenbergEffect:
Great. I will get this merged shortly.

Issue #134, opened 2015-06-23
Error using viewPorts and child router


ceoaliongroo:
Hi,  

I have a router with viewPorts:
__app.js__ 
```javascript
  ...
  config.map([
      {
        route: ['', 'home'],
        name:'home',
        viewPorts:{
          content: {
            moduleId: 'dashboard/cockpit'
          },
          menu: {
            moduleId: 'menu/menu-options'
          }
        },
        title:'Home'}
    ]);
 ...
```
app.html
```html
   ...
  <div class="page-host menu">
    <router-view name="menu"></router-view>
  </div>
  <div class="page-host content">
    <router-view name="content"></router-view>
  </div>
  ...
```

and a child route:
__cockpit.js__
```javascript
    ...
    config.map([
      {
        route: ['', 'subscribers'],
        name:'subscribers',
        viewPorts: {
          test: {
            moduleId: 'dashboard/subscription/list'
          }
        },
        title:'Subscribers'
      }
    ]);
   ...
```
__cockpit.html__
```html
  ...
  <router-view name="test"></router-view>
  ...
```

I receive the error 
```javascript
There was no router-view found in the view for dashboard/subscription/list
```

if you want to check [all the code](https://github.com/ceoaliongroo/sandbox-aurelia/tree/issue-router-viewPorts%2BchildRoute)

bryanrsmith:
@ceoaliongroo I'm not sure I understand the problem. I can't seem to get any error to appear by using your sample app. Also, it looks like your dashboard/subscription/list.html has a `<router-view>` element, but the view model in list.js does not configure a router that uses it.

Could you tell me the steps to reproduce the error in your app?

ceoaliongroo:
Yes, i think i had an error in dependencies, sorry. 
I'll fixed  https://github.com/ceoaliongroo/sandbox-aurelia/commit/fe6b6dc5b3a05fed95d3394a84d856a6041f25f0

try, again with:

```bash
git checkout issue-router-viewPorts+childRoute
git pull
jspm install -y
```
You could see something like this:
![screen shot 2015-06-29 at 10 23 26 am](https://cloud.githubusercontent.com/assets/2622152/8403031/d351f83e-1e49-11e5-91e0-c1521b5a7c5a.png)
![screen shot 2015-06-29 at 10 24 28 am](https://cloud.githubusercontent.com/assets/2622152/8403075/066a5b1c-1e4a-11e5-8407-6e8e3743c9a6.png)

PR #135, opened 2015-06-23
WIP: Force continue if 'router.viewPorts' is empty.

ceoaliongroo:
#134 

As far i understood, the ``child router`` is commited for each ``viewPort`` defined in the parent router. 

The error is generated when don't found `route-view`  defined in one of the view port (in this case ``menu-options``)

I'll expect only expected the ``child route`` only will commit on the view port defined.

Now i make a bypass of the exception, when the viewPorts are empty.

@bryanrsmith, @EisenbergEffect could you help me, to understand better about (relation between router, view (viewPorts), navigation) to found source of the problem to improve the fix?   

thanks

### DIFF
--- a/src/navigation-context.js
+++ b/src/navigation-context.js
@@ -54,7 +54,10 @@ export class NavigationContext {
       var viewPort = router.viewPorts[viewPortName];
 
       if(!viewPort){
-        throw new Error(`There was no router-view found in the view for ${viewPortInstruction.moduleId}.`);
+        if (!Object.keys(router.viewPorts).length) {
+          return;
+        }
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
@bryanrsmith Any thoughts on this?

bryanrsmith:
Sorry; I told @ceoaliongroo I'd try to write up some brief info on how viewports work to help with this, and I haven't got to it yet. I'll try to look into the problem a bit more this weekend and provide whatever info I can to help.

ceoaliongroo:
Thanks for the effort, i keep looking depth the code.

bryanrsmith:
I'm going to go ahead and close this because there hasn't been any movement. Please feel free to reopen or let me know if this is an error or if there's more I can do to help.

PR #143, opened 2015-07-03
chore(all): add types

JeroenVinke:
Please check if I specified the optional/required parameters correctly.

I'm also not sure what the types of `evaluator` and `context`should be on `evaluateNavigationStrategy()`

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "aurelia-router",
-  "version": "0.9.0",
+  "version": "0.10.0",
   "description": "A powerful client-side router.",
   "keywords": [
     "aurelia",
@@ -16,11 +16,12 @@
     "url": "http://github.com/aurelia/router"
   },
   "dependencies": {
-    "aurelia-dependency-injection": "^0.8.1",
-    "aurelia-event-aggregator": "^0.5.0",
-    "aurelia-history": "^0.5.0",
-    "aurelia-path": "^0.7.0",
-    "aurelia-route-recognizer": "^0.5.0",
+    "aurelia-dependency-injection": "^0.9.0",
+    "aurelia-event-aggregator": "^0.6.0",
+    "aurelia-history": "^0.6.0",
+    "aurelia-logging": "^0.6.0",
+    "aurelia-path": "^0.8.0",
+    "aurelia-route-recognizer": "^0.6.0",
     "core-js": "zloirock/core-js"
   }
 }
--- a/config.js
+++ b/config.js
@@ -16,38 +16,42 @@ System.config({
 
 System.config({
   "map": {
-    "aurelia-dependency-injection": "github:aurelia/dependency-injection@0.8.1",
-    "aurelia-event-aggregator": "github:aurelia/event-aggregator@0.5.0",
-    "aurelia-history": "github:aurelia/history@0.5.0",
-    "aurelia-logging": "github:aurelia/logging@0.5.0",
-    "aurelia-path": "github:aurelia/path@0.7.0",
-    "aurelia-route-recognizer": "github:aurelia/route-recognizer@0.5.0",
+    "aurelia-dependency-injection": "github:aurelia/dependency-injection@0.9.0",
+    "aurelia-event-aggregator": "github:aurelia/event-aggregator@0.6.0",
+    "aurelia-history": "github:aurelia/history@0.6.0",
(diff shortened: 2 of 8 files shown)


src/router.js
```diff
@@ -25,7 +25,7 @@ export class Router {
     return false;
   }
 
-  registerViewPort(viewPort, name) {
+  registerViewPort(viewPort:any, name?:string) {
```
bryanrsmith: We'll eventually need an interface for view ports, but in the meantime `Object` would be more appropriate than `any`.

src/router.js
```diff
@@ -46,7 +46,7 @@ export class Router {
     }
   }
 
-  configure(callbackOrConfig) {
+  configure(callbackOrConfig:any):Router {
```
bryanrsmith: `callbackOrConfig` is a `RouterConfiguration` or a callback that takes a `RouterConfiguration`.

src/router.js
```diff
@@ -60,7 +60,7 @@ export class Router {
     return this;
   }
 
-  navigate(fragment, options) {
+  navigate(fragment?:string, options?:Object):boolean {
```
bryanrsmith: I don't believe `fragment` is optional.

src/router.js
```diff
@@ -77,13 +77,13 @@ export class Router {
     this.history.navigateBack();
   }
 
-  createChild(container) {
+  createChild(container?:Router):Router {
```
bryanrsmith: `container` is a `Container` from aurelia/dependency-injection.

src/router.js
```diff
@@ -77,13 +77,13 @@ export class Router {
...
     var childRouter = new Router(container || this.container.createChild(), this.history);
     childRouter.parent = this;
     return childRouter;
   }
 
-  createNavigationInstruction(url = '', parentInstruction = null) {
+  createNavigationInstruction(url:string = '', parentInstruction?:NavigationInstruction = null):Promise<any> {
```
bryanrsmith: This returns `Promise<NavigationInstruction>`

src/router.js
```diff
@@ -133,12 +133,12 @@ export class Router {
...
   }
 
-  createNavigationContext(instruction) {
+  createNavigationContext(instruction:NavigationInstruction):NavigationContext {
     instruction.navigationContext = new NavigationContext(this, instruction);
     return instruction.navigationContext;
   }
 
-  generate(name, params) {
+  generate(name:string, params?:Object) {
```
bryanrsmith: This returns a string.

src/router.js
```diff
@@ -147,7 +147,7 @@ export class Router {
     return createRootedPath(path, this.baseUrl, this.history._hasPushState); 
   }
 
-  createNavModel(config) {
+  createNavModel(config:NavModel):NavModel {
```
bryanrsmith: `config` is a route config here, which we'll need an interface for. In the meantime, `Object` would be better.

src/router.js
```diff
@@ -215,15 +215,15 @@ export class Router {
...
     return !!(this.recognizer.hasRoute(name) || this.parent && this.parent.hasRoute(name));
   }
 
-  hasOwnRoute(name) {
+  hasOwnRoute(name:string):boolean {
     return this.recognizer.hasRoute(name);
   }
 
-  handleUnknownRoutes(config) {
+  handleUnknownRoutes(config?:string) {
```
bryanrsmith: `config` is a string, object, or function here.

src/router.js
```diff
@@ -281,7 +281,7 @@ function validateRouteConfig(config) {
   }
 }
 
-function evaluateNavigationStrategy(instruction, evaluator, context){
+function evaluateNavigationStrategy(instruction:NavigationInstruction, evaluator, context): NavigationInstruction {
```
bryanrsmith: This returns a `Promise<NavigationInstruction>`

bryanrsmith: I believe `evaluator` is a function, and `context` is a route config object (`Object` is fine for now).

bryanrsmith:
Thanks for working on this! I think there are just a few types to adjust, but it's looking good.

We don't update files in dist until we publish releases, so it'd be great if you could back out the changes there and only include the src files in your PR.

JeroenVinke:
Made a new PR with these changes: https://github.com/aurelia/router/pull/145

PR #147, opened 2015-07-04
Lists all properties on the class directly.

### LINKED ISSUE #146: RouterConfiguration in aurelia-router.d.ts is missing properties.
Using TypeScript, I have a function like
```ts
public configureRouter(config: RouterConfiguration, router: Router) {
	config.title = 'foo';
}
```

The TypeScript compiler errors on `config.title` because the `RouterConfiguration` class doesn't have a `title` property on it.  In https://github.com/aurelia/router/blob/master/src/router-configuration.js#L63 we can see that title is an optional parameter, so it should be included in the d.ts as such.

I'm guessing that the d.ts auto-generator doesn't have the ability to see properties, so perhaps the right solution is to just add a set method like `setTitle(title)`.  This way, those using TypeScript can get some amount of static typing and don't have to resort to using `any` for their `configureRouter` method.

On a related note, ES6 supports the TypeScript type syntax without error.  For example, you could write an ES6 function that looks like this:
```js
exportToRouter(router: Router) {
	...
}
```
And it will compile just fine in ES6, with the type information simply ignored.  Perhaps in order to get better TypeScript typings the Aurelia ES6 source code could be annotated with type information where it is easy/reasonable.  Then the auto-generated d.ts files could have extended type information.

Alternatively, Aurelia could simply be migrated to TypeScript source base and transpiled to ES6.  ;)  I'm personally a big fan of this, though I'm guessing ES6 source was an intentional choice.

Zoltu:
Fixes #146.

I'm not sure if empty properties will make it to the d.ts or not (they are compiled out when transpiling to ES5 I believe).  I could default them to null, though I'm worried that some area of the code may treat null and undefined differently.  Alternatively, #146 could be fixed by adding a setter for `title` (`unknownRouteConfig` already has one).  I prefer this method, as it makes it clear at the top of the class what all the properties of it are, but perhaps people shouldn't be reading from these, only setting them in which case the public interface is really set-only, so a function for each would be better.

### DIFF
--- a/src/router-configuration.js
+++ b/src/router-configuration.js
@@ -1,11 +1,11 @@
 import {RouteFilterContainer} from './route-filters';
 
 export class RouterConfiguration{
-  constructor() {
-    this.instructions = [];
-    this.options = {};
-    this.pipelineSteps = [];
-  }
+  instructions = [];
+  options = {};
+  pipelineSteps = [];
+  title;
+  unknownRouteConfig;
 
   addPipelineStep(name, step) {
     this.pipelineSteps.push({name, step});


bryanrsmith:
Looks good to me! @Zoltu, would you mind updating the commit message to conform to [Aurelia's guideline](https://github.com/DurandalProject/about/blob/master/CONTRIBUTING.md#commit)? Something like `refactor(RouterConfiguration): list all properties on the class directly`. This format makes it easier to quickly compile change logs and such.

Zoltu:
Done.

PR #163, opened 2015-07-20
Added a "commitcomplete" pipeline step

andy-scott:
Add a Pipeline Step to allow code to run after the View has been
inserted into the DOM.

We use some jquery plugins that cache elements against a global object
to save editing these plugins which makes updates difficult, we use the
commitcomplete pipline step to reset the values on the global object


bryanrsmith:
Sounds reasonable. I think it'd be nice to name it "postcommit", for consistency with the existing "precommit" step.

andy-scott:
I have updated the name to "postcomplete".

PR #171, opened 2015-07-26
chore(Router): added currentInstruction property

JeroenVinke:
I have added the `currentInstruction` property to the `Router` class, so that it is added to the generated d.ts file.


EisenbergEffect:
@bryanrsmith See any issues with this? Looks ok to merge to me.

bryanrsmith:
:+1:

PR #183, opened 2015-08-15
Added types and improved CS

nomack84:
I've added types to several types and fixed some CS issues.

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -58,7 +58,7 @@ export class Router {
   * @param container The [[Container]] to use when child routers.
   * @param history The [[History]] implementation to delegate navigation requests to.
   */
-  constructor(container, history) {
+  constructor(container : Container, history : History) {
     this.container = container;
     this.history = history;
     this._configuredPromise = new Promise((resolve => {
@@ -68,9 +68,9 @@ export class Router {
   }
 
   /**
-  * Gets a valid undicating whether or not this [[Router]] is the root in the router tree. I.e., it has no parent.
+  * Gets a valid indicating whether or not this [[Router]] is the root in the router tree. I.e., it has no parent.
   */
-  get isRoot() {
+  get isRoot() : boolean {
     return false;
   }
 
@@ -80,7 +80,7 @@ export class Router {
   * @param viewPort The viewPort.
   * @param name The name of the viewPort. 'default' if unspecified.
   */
-  registerViewPort(viewPort:Object, name?:string) {
+  registerViewPort(viewPort : Object, name? : string) : void {
     name = name || 'default';
     this.viewPorts[name] = viewPort;
   }
@@ -88,7 +88,7 @@ export class Router {
   /**
   * Returns a Promise that resolves when the router is configured.
   */
-  ensureConfigured():Promise<void> {
+  ensureConfigured() : Promise<void> {
     return this._configuredPromise;
   }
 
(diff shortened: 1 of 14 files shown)


src/router.js
```diff
@@ -80,15 +80,14 @@ export class Router {
   * @param viewPort The viewPort.
   * @param name The name of the viewPort. 'default' if unspecified.
   */
-  registerViewPort(viewPort:Object, name?:string) {
-    name = name || 'default';
+  registerViewPort(viewPort : Object, name? : string = 'default') : void {
```
bryanrsmith: I'm a little nervous about this change. I think templating-router may end up passing empty string to this, and this change would break that. Have you checked out existing callers of this method?

bryanrsmith:
Thanks, @nomack84. Would you mind updating your commit message to match our contributing guidelines?

These changes look great. If you're interested in helping us improve code quality, we're just starting to roll out more comprehensive style and linter config. You can check out a PR that makes that change in the metadata repo here: aurelia/metadata#20 (The PR is out of date now, but the config changes and such are still valid).

src/router.js
nomack84: I didn't think about the possibility of empty strings, so this indeed will break the code. I'll revert this change.

nomack84:
@bryanrsmith I love a lot Aurelia and admire a lot the great work you guys are doing. Even with such a great difference in resources and people, you are doing much better than Angular, at least that's what I think. So definitely if I can do at least something as simple as improving code quality, certainly can count on me.

nomack84:
By the way, where can I find the contributing guidelines?

nomack84:
Forgot it, I've found them :)

bryanrsmith:
Thank you for the kind words. We would of course appreciate any contributions you might have time for.

https://github.com/DurandalProject/about/blob/master/CONTRIBUTING.md#commit

src/router.js
nomack84: It doesn't break it, because `default` is assumed only when `undefined` is passed. You can check it out. 

```javascript
function defaults(x = 'default') {
  console.log(x);
}

defaults(''); -> prints ''
defaults('view'); -> prints 'view'
defaults(); -> prints 'default'
```
Check here: http://www.es6fiddle.net/idjr4tkv/

nomack84: What could be a valid argument against default params is the code Babel generates.
```javascript 
function defaults(x = 'default') {
  console.log(x);
}
```
generates:
```javascript
function defaults() {
  var x = arguments.length <= 0 || arguments[0] === undefined ? 'default' : arguments[0];

  console.log(x);
}
```
Looking at this, perhaps we is better to use the 'old' way. What do you think?

bryanrsmith: I don't mind the generated code, I just want to make sure that the code in aurelia/templating-router never passes null or empty string and expects the default to be applied. I am not sure what how behavior will change in the cases of `<router-view></router-view>` and `<router-view name=""></router-view>`.

nomack84: Well, I checked the behavior in the above fiddle and it doesn't fail. If an empty string is passed, it doesn't use the `default`. So I don't see why it'll fail. Is there any test that cover this, or perhaps I should add it...

EisenbergEffect:
CLA confirmed. @bryanrsmith Proceed as you deem appropriate.

src/router.js
bryanrsmith: Right... viewPorts shouldn't be allowed to have an empty string for a name. If empty string is passed, it should get the default.

bryanrsmith:
@nomack84 I think we should revert the default value change. Other than that this looks good to me. Could you please rebase against master, squash your commits, and update the commit message?

nomack84:
@bryanrsmith I've reverted the change and updated the commit message. Can you help me rebasing against master? Any guidance about how to achive it?
Thanks!

bryanrsmith:
Sure. Here's a good explanation with steps to take: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

Let me know if I can help further.

nomack84:
@bryanrsmith I've rebased against master already.

bryanrsmith:
It looks like it's still out of date. I was able to resolve the conflicts myself, but unfortunately the circular dependencies added in this PR (for type annotations) is preventing it from building (`gulp build`). We use a tool that calculates concatenation order based on import statements, and it doesn't handle circular dependencies at the moment. I'll try to take a look at that this weekend. I won't be able to merge this until I've got that fixed.

bryanrsmith:
@nomack84 Nevermind. I was able to get it working by removing the imports that were only needed for type annotations. I went ahead and resolved the conflicts myself and pushed the change to master -- 8e528c858c8aacc5a5f8c7f087516de1e160d5d9

Thanks for the contribution!

nomack84:
Happy to help :)

PR #192, opened 2015-08-24
chore(type annotation): changed type info

behzad888:


bryanrsmith:
Thanks @behzad888. The .d.ts file is generated from the type annotations in the source, so unfortunately I'm unable to merge this. Would you be willing to port these changes back to the src files?

behzad888:
Yes, I will change that.

PR #194, opened 2015-08-26
fix(activation.js) fix failing tests

nomack84:
This fix a typo in the `activation.js` file when is used `if` instead of `for`.

### DIFF
--- a/src/activation.js
+++ b/src/activation.js
@@ -70,7 +70,7 @@ function findDeactivatable(plan, callbackName, list) {
 
       var controller = prevComponent.bindingContext;
 
-      if (callbackName in controller) {
+      for (callbackName in controller) {
         list.push(controller);
       }
     }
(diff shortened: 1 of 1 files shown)


bryanrsmith:
Can you explain the problem this solves? The previous version looks correct to me.

nomack84:
Tests... run the tests on the master and you'll see

nomack84:
Well... I check it carefully and it turns out that it was failing because the controllers were undefined, I update my PR. Please check it out.

bryanrsmith:
@EisenbergEffect Could you review this one? I'm not up to speed on the templating changes. Is this just happening because the tests are running against the current release of templating instead of master?

EisenbergEffect:
I'll take care of this one.

EisenbergEffect:
This doesn't seem right to me. If there's a bug on master it probably relates to the changes that were made to the templating engine's internal structures that I'm trying to lock down. I think an update of the deps and a test run would probably see them pass.

PR #196, opened 2015-08-28

behzad888:


EisenbergEffect:
Adding any types isn't necessary. That happens by default. We do have more precise types for these apis which we can add soon. Closing for now.

PR #197, opened 2015-08-28

behzad888:


EisenbergEffect:
Types here aren't exactly correct. title should be a string. Closing for now.

PR #198, opened 2015-08-28

behzad888:


EisenbergEffect:
Any types aren't necessary and we've got more specific types available we could use. Closing for now.

PR #199, opened 2015-08-28

behzad888:


EisenbergEffect:
Any types not necessary. More specific types could be added.

PR #200, opened 2015-08-28

behzad888:


nomack84:
@behzad888 did you see https://github.com/aurelia/router/pull/183

EisenbergEffect:
Adding any types isn't necessary. There are more accurate types that could be added.

PR #202, opened 2015-08-29
chore(all): add eslint and fix lint errors

bryanrsmith:


IanVS:
Looks like you've already taken care of your lint errors, but in case you have any other projects you want to add ESLint to, I made a tool, [ESLint Nibble](https://github.com/IanVS/eslint-nibble) to try to make it a little easier, by getting an overview of the rules that are failing, and then focusing on one rule at a time.  If you run into any trouble with ESLint, feel free to give us a shout.

PR #205, opened 2015-09-01
Typescript definition fix

### LINKED ISSUE #204: Typescript def: mapUnknownRoutes
If `config` is a function, it returning a promise is optional.

jods4:
Fixes #204


EisenbergEffect:
Good catch!

jods4:
Easy catch, my TS build broke ;)

PR #219, opened 2015-09-29
fix(router): properly handle async router configurations

### LINKED ISSUE #213: Multiple viewports cause duplicate routes
I'm trying to setup multiple viewports with Aurelia (http://aurelia.io/docs.html#rendering-multiple-view-ports). However, when I setup the router config to tell which ```<router-view>``` which module to load, the result is that now, ```router.navigation``` has an additional entry for each viewport added.  

For example (modified from the skeleton project):
``` javascript
configureRouter(config, router) {
        config.map([
      { route: ['','welcome'],  name: 'welcome',      viewPorts: {test: {moduleId: 'welcome'}, 		test2: {moduleId: 'welcome'}},      nav: true, title:'Welcome' },
      { route: 'users',         name: 'users',        viewPorts: {test: {moduleId: 'users'}, 		test2: {moduleId: 'users'}},        nav: true, title:'Github Users' }
    ]);
        this.router = router;
}
```
``` html
<div class="page-host">
    <router-view name="test"></router-view>
    
    <router-view name="test2"></router-view>
  </div>
```

The result is now ```router.navigation``` contains duplicate navigations (one duplicate for each router-view)
screen shot 2015-09-18 at 9 52 27 am
screen shot 2015-09-18 at 9 53 33 am

Any ideas on this one? Could it be bad configuration? Or is it something wrong with the router?...

bryanrsmith:
This change moves handling of async router config callbacks into Router.configure to resolve a race condition when using multiple viewPorts. Async configuration was previously handled externally in app-router and route-loading, but that code failed to block subsequent calls from configuring multiple times. Router.configure() now handles async callbacks, setting isConfigured immeditely to allow callers to check state.

BREAKING CHANGES:
  * Router.configure() now always returns a Promise, and is no longer chainable.
  * Router.configure() now applies the configuration in a Promise continuation, so it is not safe to assume that the router will be configured synchronously when the configuration callback is synchronous.

fixes #213

### DIFF
--- a/src/app-router.js
+++ b/src/app-router.js
@@ -6,7 +6,6 @@ import {Router} from './router';
 import {PipelineProvider} from './pipeline-provider';
 import {isNavigationCommand} from './navigation-commands';
 import {EventAggregator} from 'aurelia-event-aggregator';
-import {RouterConfiguration} from './router-configuration';
 
 const logger = LogManager.getLogger('app-router');
 
@@ -85,21 +84,19 @@ export class AppRouter extends Router {
 
     if (!this.isActive) {
       let viewModel = this._findViewModel(viewPort);
-
       if ('configureRouter' in viewModel) {
-        let config = new RouterConfiguration();
-        let result = Promise.resolve(viewModel.configureRouter(config, this));
-
-        return result.then(() => {
-          this.configure(config);
-          this.activate();
-        });
+        if (!this.isConfigured) {
+          return this.configure(config => viewModel.configureRouter(config, this))
+            .then(() => { this.activate(); });
+        }
+      } else {
+        this.activate();
       }
-
-      this.activate();
     } else {
       this.dequeueInstruction();
     }
+
+    return Promise.resolve();
   }
 
   _findViewModel(viewPort: Object) {
--- a/src/router.js
+++ b/src/router.js
@@ -91,23 +91,22 @@ export class Router {
   * Configures the router.
   *
   * @param callbackOrConfig The [[RouterConfiguration]] or a callback that takes a [[RouterConfiguration]].
-  * @chainable
   */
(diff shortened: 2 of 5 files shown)


src/app-router.js
```diff
@@ -85,21 +84,19 @@ export class AppRouter extends Router {
...
+      } else {
+        this.activate();
       }
-
-      this.activate();
     } else {
       this.dequeueInstruction();
     }
+
+    return Promise.resolve();
```
bryanrsmith: This method now always returns a `Promise`.

src/router.js
```diff
@@ -91,23 +91,22 @@ export class Router {
...
+      config = new RouterConfiguration();
+      result = callbackOrConfig(config);
     }
 
-    this.isConfigured = true;
-    this._resolveConfiguredPromise();
-
-    return this;
+    return Promise.resolve(result).then((c) => {
+      (c || config).exportToRouter(this);
```
bryanrsmith: A side effect of this refactor is that the configuration callback may return a new `RouterConfiguration`, if it wants.

test/router.spec.js
```diff
@@ -66,172 +66,166 @@ describe('the router', () => {
   });
 
   describe('generate', () => {
-    it('should generate route URIs', () => {
+    it('should generate route URIs', (done) => {
```
bryanrsmith: There are two new tests. All other changes to tests are just moving assertions into `Promise` continuations to handle the new async `Router.configure` implementation.

EisenbergEffect:
@bryanrsmith Sounds good. Feel free to merge it in and continue on :) Thanks!

AStoker:
Any update on when this will be included in the release? There are a few projects on hold (talked on the Aurelia Gitter) for the next release with it's bug fixes.

PR #223, opened 2015-10-07
feat(activation): remove support for string affirmations from activation

bryanrsmith:
BREAKING CHANGE: Activation code returning string values to abort
activation will need to perform their own comparison and return a
boolean


EisenbergEffect:
Probably a good idea..and probably nobody was using this. Please feel free to merge and move ahead.

PR #235, opened 2015-10-28
Improve handling of unknown route configs

bryanrsmith:


src/router.js
```diff
@@ -286,28 +285,17 @@ export class Router {
...
   handleUnknownRoutes(config?: string|Function|RouteConfig): void {
-    let callback = instruction => new Promise((resolve, reject) => {
-      function done(inst) {
-        inst = inst || instruction;
-        inst.config.route = inst.params.path;
-        resolve(inst);
-      }
-
-      if (!config) {
-        instruction.config.moduleId = instruction.fragment;
```
bryanrsmith: This behavior seems highly specialized, and likely to cause very confusing problems if null or undefined are accidentally passed. It's trivial for an application to implement if desired, so I think it's better not to include it by default.

EisenbergEffect:
I agree with you. I like it. Please go ahead and merge @bryanrsmith

StrahilKazlachev:
@bryanrsmith Can you give some details what properties the returned `RouteConfig` can have? Setting the `route` property causes `ERROR [app-router] Error: Invalid Route Config: You must specify a route pattern`. I was trying to make the router populate the `params` property of the `NavigationInstruction`, to make it look like the one of a registered route.

bryanrsmith:
@StrahilKazlachev You will need to set a moduleId or redirect. For unknown routes like this, the `route` property will always be set to the incoming path, and there's currently no API for parsing it and providing params. That's probably something we should add, though. I'll open a new issue for that.

StrahilKazlachev:
Ok, so what are the relevant properties then? So far I've used `moduleId` and `title` and it worked as expected. You said that `redirect` is also possible to set. `nav`, `href`, `name` are probably irrelevant in this scenario. What about `navigationStrategy` and `viewPorts`? Also should `handleUnknownRoutes` return `RouteConfig`, since it's only mandatory property is `route`, or some new type(just a thought)?

PR #244, opened 2015-11-09
fix(updateTitle): use private method name

jdanyow:


EisenbergEffect:
@bryanrsmith Can I get a quick confirmation from you that this is good to merge?

bryanrsmith:
Thanks, good catch!

EisenbergEffect:
Thanks!

PR #260, opened 2015-12-11
chore(ci): configure jspm github auth

bryanrsmith:


EisenbergEffect:
Any reason not to merge?

bryanrsmith:
It's not quite ready. I had previously added `jspm install` to the `npm postinstall` script to simplify installation and let us avoid installing jspm globally. That won't work with this change because jspm needs to be installed before we configure it. I need to swap that around now. I'll try to wrap it up tonight.

PR #272, opened 2016-01-05
fix(app-router.js): restore queue object reference

davismj:


EisenbergEffect:
@davismj The tests are failing. I know why. The initial value of the array is never being set. If you can amend your commit to check whether the array exists first, then set it if not or reset it if it does, that should fix it.

PR #273, opened 2016-01-05
fix(app-router): restore _queue object reference

davismj:


src/app-router.js
```diff
@@ -28,7 +28,11 @@ export class AppRouter extends Router {
   reset() {
     super.reset();
     this.maxInstructionCount = 10;
-    this._queue = [];
+    if (!this._queue) {
+      this._queue = [];
```
bryanrsmith: If the property is never reassigned we may as well initialize it inline and skip the check here.

EisenbergEffect: Well, if it actually being reset, it will be assigned. The constructor also runs "reset" to set everything to the initial state. That's when the queue would get initially set. It's a little ugly, but it works.

davismj: I thought the same thing @bryanrsmith

EisenbergEffect: I think the inline assignment could be problematic. It depends on the code-gen. If the super constructor runs before the inline initialization, then I think you could get an error because the super constructor calls reset and that would attempt to set the length before the queue was initialized. Does that make sense?

davismj: Oh, yea I think so. That's why I submitted it this way.

bryanrsmith: Eew. Nevermind.

EisenbergEffect: We could refactor it. I was trying to not duplicate the code though. What do you guys think?

davismj: I could refactor it. I don't like putting initialization code in a reset function. Is that okay?

EisenbergEffect: It's fine. Go ahead. You can just amend your PR with that improvement.

davismj: On second thought, I don't think it's going to get any better than it is now.

EisenbergEffect: Ok, we can take this. It will work :)

PR #291, opened 2016-02-13
fix(router): log helpful error on missing viewport

### LINKED ISSUE #241: Crash when navigating to route with misconfigured viewPorts
Configure a view with two viewports and a router with one route that properly configures both viewports and one route that configures only one viewport. Attempting to navigate from the good route to the bad will crash in buildNavigationPlan. We should log a helpful error message.

jedd-ahyoung:
When a route does not configure all available viewports, the navigation plan will crash. Now, the navigation plan will log a helpful error with the missing viewport and the affected route.

Fixes #241.


bryanrsmith:
Looks good! I think it just needs some lint cleanup. `gulp lint` should show you the problems locally.

jedd-ahyoung:
Yup, I forgot to run that. I've fixed the issue and force pushed the branch. Let me know if there are other issues (specifically, the `if (!value)` instead of the explicit check `if (value === null || value === undefined)`.

PR #294, opened 2016-02-19
feat(router): generate absolute URI from router

### LINKED ISSUE #88: Support scheme, host, and port in route generation
It would be great to be able to pass a configuration object to `router.generate` that would enable fully qualified URIs to be rendered. As it is now, you only get a relative path (which is fine by default).

jedd-ahyoung:
Adds an extra boolean argument to `router.generate` to specify whether or not to generate an absolute URI.

Fixes #88.

### DIFF
--- a/config.js
+++ b/config.js
@@ -22,7 +22,7 @@ System.config({
     "aurelia-route-recognizer": "npm:aurelia-route-recognizer@1.0.0-beta.1.1.0",
     "babel": "npm:babel-core@5.8.35",
     "babel-runtime": "npm:babel-runtime@5.8.35",
-    "core-js": "npm:core-js@2.0.3",
+    "core-js": "npm:core-js@2.1.0",
     "github:jspm/nodelibs-assert@0.1.0": {
       "assert": "npm:assert@1.3.0"
     },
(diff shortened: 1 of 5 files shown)


jedd-ahyoung:
Note - this depends on changes in history (https://github.com/aurelia/history/pull/7) and history-browser (https://github.com/aurelia/history-browser/pull/22).

src/router.js
```diff
@@ -186,7 +186,7 @@ export class Router {
   * @param params The route params to be used to populate the route pattern.
   * @returns {string} A string containing the generated URL fragment.
   */
-  generate(name: string, params?: any): string {
+  generate(name: string, params?: any, absolute?: boolean = false): string {
```
bryanrsmith: What do you think about using an options object for the third parameter, in case we ever need to add more options?

src/util.js
```diff
@@ -1,12 +1,16 @@
...
 
+  if (hasPushState && absolute) {
+    path = path.substring(1, path.length);
+  }
+
   return path;
 }
 
-export function _createRootedPath(fragment, baseUrl, hasPushState) {
+export function _createRootedPath(fragment, baseUrl, hasPushState, absolute) {
```
bryanrsmith: It seems like we should be able to eliminate the `hasPushState` hacks now that we know the base URL, right?

test/setup.js
```diff
@@ -0,0 +1,2 @@
+import {initialize} from 'aurelia-pal-browser';
+initialize();
```
bryanrsmith: Why is this necessary? I don't think this library depends on PAL at all.

config.js
```diff
@@ -42,23 +43,27 @@ System.config({
...
     "npm:aurelia-route-recognizer@1.0.0-beta.1.1.0": {
       "aurelia-path": "npm:aurelia-path@1.0.0-beta.1.1.0",
-      "core-js": "npm:core-js@2.0.3"
+      "core-js": "npm:core-js@2.1.0"
     },
     "npm:babel-runtime@5.8.35": {
       "process": "github:jspm/nodelibs-process@0.1.2"
     },
-    "npm:core-js@2.0.3": {
+    "npm:core-js@2.1.0": {
```
bryanrsmith: Does this need to be coordinated with the version of core-js in use by other aurelia libs? Or maybe it doesn't matter if it's being removed. @EisenbergEffect

nomack84: @EisenbergEffect already started to remove `core-js` hard dependency, perhaps is better to keep current version and in a different PR remove `core-js` and if needed, add the dependency on `aurelia-polyfills` library

src/router.js
nomack84: is definitely a good idea to use an option object, it'll make keeping BC changes easier

src/util.js
jedd-ahyoung: I'm not sure if this is possible. The `hasPushState` argument actually determines whether or not we add the hash character to a generated URL - it's only passed to `_createdRootedPath` to get passed to the normalize function. I'm pretty sure that's necessary, and there's no way that having the absolute root URL would let us know whether or not to use the hash.

src/router.js
jedd-ahyoung: Agreed, it's more extensible.

EisenbergEffect: Sounds good to me as well. Let's go with the options object.

src/util.js
bryanrsmith: The hash character should be part of the history's base URL. The router only cares about the path fragment, and the hash is really an implementation detail of the history.

jedd-ahyoung: Perhaps you're right about that - the separation of concerns does make sense - but that would then mean that relative routes would still need to add a hash, or generate the full URL and then extract the desired section. It also means that generating a link to an empty route ('') would always have the hash included. Right now, the default generate function creates a relative path. Do we intend to change that?

test/setup.js
jedd-ahyoung: That's an oversight on my part. I'll remove it.

EisenbergEffect:
I'm removing core-js as a hard dependency and using our own aurelia-polyfills as a dev dependency when they are needed for tests. This will let the developer decide what they want in the end, with a favor towards our own which will be the minimum requirement...allowing it to be smaller and faster.

Keep up the good work guys. Let's get this and the other two related improvements wrapped up and in as soon as we can.

PR #296, opened 2016-02-24
refactor(router): rename pipeline slots

### LINKED ISSUE #295: Navigation Pipeline as normal middleware pipeline rather than specific hooks
So I've just used the `config.addPipelineStep('authorize', AuthRouterPipelineStep);` in the `configureRouter` function.

I initially tried to do:

```config.addPipelineStep('authorise', AuthRouterPipelineStep);```

That didn't work, and it was until I changed the name to `authorize` till it worked. This seem very unintuative to me. I was informed there are different hooks in the navigation pipeline, and that `authorize` was a specific hook.

I don't fully understand these hooks. It would seem better to me that you have a normal middleware pipeline. Named, or unnamed. 

I come from a C# background using WebApi and ASPNET core where the middleware have an injectable constructor and injectable invoke function. The Aurelia pipeline steps follow the same pattern. They seem to slot in in exactly the same way. This is probably what led to my confusion because they don't actually work in the same way at all. 

Now I do see the point in the hooks, but the current way seems weird to me.

It would seem to me that it would be better to do something like this:

```
config.insertIntoAuthoriseStep(AuthRouterPipelineStep)
config.insertIntoModelBindStep(SomeBindPipelineStep)
```

Here it's clear at what stage you want to hook into, and a little bit of documentation saying that you can have multiple middleware inserted into each step makes so much more sense to me.

I don't know if this is way out from what you guys were planning with this, or if I'm way off the mark here.

But either way, the fact that there are these "magic' strings defining navigation hooks is not ideal.

### PR DESCRIPTION
**Pull request by jedd-ahyoung**
Pipeline slots have been renamed: 'modelbind' is 'preActivate', 'precommit' is 'preRender', and 'postcomplete' is 'postRender'. Sugar methods have been added to support the new names. RouteFilterContainers now support aliases in order to preserve backwards compatibility. Tests have been added for 'route-filters.js'.

Fixes #295.

### DIFF
--- a/src/pipeline-provider.js
+++ b/src/pipeline-provider.js
@@ -23,15 +23,15 @@ export class PipelineProvider {
       BuildNavigationPlanStep,
       CanDeactivatePreviousStep, //optional
       LoadRouteStep,
-      createRouteFilterStep('authorize'),
-      createRouteFilterStep('modelbind'),
+      createRouteFilterStep(pipelineSlot.authorize),
       CanActivateNextStep, //optional
+      createRouteFilterStep(pipelineSlot.preActivate, { aliases: ['modelbind']}),
       //NOTE: app state changes start below - point of no return
       DeactivatePreviousStep, //optional
       ActivateNextStep, //optional
-      createRouteFilterStep('precommit'),
+      createRouteFilterStep(pipelineSlot.preRender, { aliases: ['precommit']}),
       CommitChangesStep,
-      createRouteFilterStep('postcomplete')
+      createRouteFilterStep(pipelineSlot.postRender, { aliases: ['postcomplete']})
     ];
   }
 
@@ -44,3 +44,10 @@ export class PipelineProvider {
     return pipeline;
   }
 }
+
+const pipelineSlot = {
+  authorize: 'authorize',
+  preActivate: 'preActivate',
+  preRender: 'preRender',
+  postRender: 'postRender'
+};
--- a/src/route-filters.js
+++ b/src/route-filters.js
@@ -5,14 +5,22 @@ export class RouteFilterContainer {
 
   constructor(container: Container) {
     this.container = container;
+    this.lookup = { };
     this.filters = { };
     this.filterCache = { };
   }
 
+  register(key: string, aliases: string[]) {
+    aliases.forEach((alias) => {
+      this.lookup[alias] = key;
+    });
+  }
+
   addStep(name: string, step: any, index: number = -1): void {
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: src/route-filters.js [hunk 1]
```diff
@@ -23,33 +31,36 @@ export class RouteFilterContainer {
...
   }
 }
 
-export function createRouteFilterStep(name: string): Function {
+export function createRouteFilterStep(name: string, options?: any = {}): Function {
+  options = Object.assign({}, { aliases: [] }, options);
   function create(routeFilterContainer) {
-    return new RouteFilterStep(name, routeFilterContainer);
+    let key = Symbol(name);
+    routeFilterContainer.register(key, [name].concat(options.aliases));
```
bryanrsmith: I think `[ name, ...options.aliases ]` reads nicely for this.

src/pipeline-provider.js
```diff
@@ -23,15 +23,15 @@ export class PipelineProvider {
       BuildNavigationPlanStep,
       CanDeactivatePreviousStep, //optional
       LoadRouteStep,
-      createRouteFilterStep('authorize'),
-      createRouteFilterStep('modelbind'),
+      createRouteFilterStep(pipelineSlot.authorize),
+      createRouteFilterStep(pipelineSlot.preActivate, { aliases: ['modelbind']}),
       CanActivateNextStep, //optional
       //NOTE: app state changes start below - point of no return
```
bryanrsmith: Were we going to move preactivate to the point of no return?

bryanrsmith:
Looks good to me!

stsje:
Pipeline slots are still being hardcoded in some places
https://github.com/aurelia/router/blob/master/src/pipeline-provider.js#L20

The pipelineSlot's have got a new name, but the keys are still the same old. It's good for backwards compatibility, but you won't be able to use what should be the new key (as string) 'preActivate'.
Shouldn't the aliases be fixing this?

jedd-ahyoung:
Line items in the pull request have been addressed, symbols have been removed, and the pipeline slot names have been fixed.

EisenbergEffect:
One more change...let's not export the pipeline step consts. Lets keep them internal and just have the methods since the step names are really an implementation detail.

jedd-ahyoung:
Done.

2016-02-24 20:45 GMT-05:00 Rob Eisenberg <notifications@github.com>:

EisenbergEffect:
Cool. Thanks! @bryanrsmith Whenever you think it's ready.

PR #305, opened 2016-03-10
feat(router): add option to compare query params

### LINKED ISSUE #268: Url Parameters are not checked in BuildNavigationPlanStep
If navigating from e.g. `/app.html#/page1` to `/app.html#/page1?param1=abc` the `no-change` view port plan is choosen in the `BuildNavigationPlanStep` function. So no view model is ever informed about the navigation.
The problem is that the function `hasDifferentParameterValues`, called from `BuildNavigationPlanStep`, compares the `params` of the current `NavigationInstruction` object to the previous `NavigationInstruction` object, but not the `queryParams`.
I think the `queryParams` should also be checked so that the `invoke-lifecycle` view port plan is choosen when navigating from `/app.html#/page1` to `/app.html#/page1?param1=abc`.
I am using `npm:aurelia-router@1.0.0-beta.1` and referring to `navigation-plan.js` of the latest commit (c6af5d1).

jedd-ahyoung:
The router would not reload the view if query parameters changed in the same route. We add this behavior to navigation-plan.js and we add the `compareQueryParams` option in order to opt in to this behavior.

Fixes #268.

### DIFF
--- a/src/navigation-instruction.js
+++ b/src/navigation-instruction.js
@@ -6,7 +6,8 @@ interface NavigationInstructionInit {
   config: RouteConfig,
   parentInstruction: NavigationInstruction,
   previousInstruction: NavigationInstruction,
-  router: Router
+  router: Router,
+  options: Object
 }
 
 export class CommitChangesStep {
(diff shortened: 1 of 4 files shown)


bryanrsmith:
:+1: LGTM

atsu85:
@jedd-ahyoung, is this documented as well or a hidden feature?

Issue #271, opened 2016-01-02
Case sensitive router


DamageLimiter:
The routing is case sensitive and I am not quite sure if it's intended to be like this and url's should not be case sensitive. 

I have this route:
```
{ route: "fertigung/waage/:id", moduleId: "fertigung/waage", name: "waage", title: "Waage", nav: false, settings: { hasSubMenu: false, isSubMenu: true, parent: 'Fertigung' }
```

and I have a link dependent on that route:
`<a href="#/fertigung/Waage/${formular.Id}" class="btn btn-white btn-xs"> Waage</a>`

As you can see I have written "Waage" with a capital "W" in the html link, which results to a redirect to the default route. Shouldn't the route itself be case insensitive?

EisenbergEffect:
@bryanrsmith I'm thinking that our pattern matching should be case-insensitive but that when we pass the parsed parameters on to the activate method, they should retain their original casing. What do you think?

bryanrsmith:
I agree.

wichert:
If this is changed I would like to have an option for the router to remain case sensitive.

davidbfrogman:
I would also be interested in the router being case insensitive.

tolemac:
I have a strange behaivor with the router tests.

I have add after `it('should navigate to named routes',` this test:

    it('should navigate to case sensitive routes', (done) => {
      const options = {};
      spyOn(history, 'navigate');

      router.configure(config => {
          config.map({ name: 'test', route: 'Test/Test', moduleId: './test' });
        })
        .then(() => {
          router.navigate('/Test/Test', options);
          expect(history.navigate).toHaveBeenCalledWith('#/Test/Test', options);
          history.navigate.calls.reset();

          router.navigate('/test/Test', options);
          expect(history.navigate).toHaveBeenCalledWith('#/test/test', options);
          history.navigate.calls.reset();

          
          done();
        });
    });

And the new test passed ... Should this fail?

davidbfrogman:
I'd love to see some movement on this feature.

tolemac:
I would like to help to implement it but have no experience with tests. Anybody can tell me how to test it in order to try implement it?

EisenbergEffect:
Take a look at some of the existing tests in the repo and see how they are put together. If you have questions, I'm sure @bryanrsmith can help you out.

tolemac:
[...]

PR #327, opened 2016-04-16
fix(router): router is case sensitive by default

tolemac:
Router is case sensitive by default, if you need to set some route as case sensitive you can add `caseSensitive: true` property to the route configuration.

in order to close https://github.com/aurelia/router/issues/271

### DIFF
--- a/src/interfaces.js
+++ b/src/interfaces.js
@@ -68,5 +68,10 @@ interface RouteConfig {
   */
   navModel?: NavModel;
 
+  /**
+  * When true is specified, this route will be case sensitive.
+  */
+  caseSensitive?: boolean;
+
   [x: string]: any;
 }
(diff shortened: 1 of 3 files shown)


tolemac:
Need https://github.com/aurelia/route-recognizer/pull/21 to pass the tests

tolemac:
Ops! I think the commit message and the pull request name is wrong. It should be "router is case *in*sensitive by default".  Like route-recognizer PR...
What can I do to resolve it?

EisenbergEffect:
@bryanrsmith I'd like to get this in with the other related commit, but looks like the CI is broken. Is it just a CI error? or an issue with the PR?

bryanrsmith:
Looks like a problem with the build. I'll try to fix it on my lunch break.

EisenbergEffect:
@tolemac Again, apologies for taking so long on this. It's in for tomorrow's release 😄

PR #333, opened 2016-04-26
Added main attribute to bower.json, which will help in creating context menu map for aurelia-webpack use with bower packages.

ranjith86:
Added main attribute to bower.json, which will help in creating context menu map for aurelia-webpack use with bower packages.


EisenbergEffect:
Thanks for the reminder. Implementing this across the repos now.

Issue #332, opened 2016-04-24
Pipeline steps are not reset when changing roots


z-brooks:
I'm using roots to separate the login/register parts from the rest.  The login root doesn't need the auth step used in the main root.

Pipeline steps added in `configureRouter` are not removed when switching roots with `setRoot`.  Any steps added will remain and will be added again when switching back.
* [skeleton repro](https://github.com/z-brooks/pipeline-repro)
* [gist.run](https://gist.run/?id=099a2fe7a6d5d82d9830e04c3f1c15b3)

Something like [this](https://github.com/aurelia/router/compare/master...z-brooks:pipeline-reset) would fix it, but I'm not sure if it's the best way.

EisenbergEffect:
Pipeline steps are not managed by the router, but by a separate pipeline provider service. I would recommend that you just reference that and remove the steps you don't want as needed.

z-brooks:
Yeah, that's fine. I'm currently using the modified fork linked in the top post. 

The concern is that when switching back and forth the steps accumulate so there are multiples of the same step in the pipeline being applied when routes change.
Updated the [gist.run](https://gist.run/?id=e0a6204fd698f5597328acc992293ce6) to illustrate. Click the `switch` button a few times and the steps will add up. Then click the link and in the console there will be output for each duplicate step in the `pipelineSlot`.
This could be worked around by setting a var indicating the steps have been added, but that seems hackish.
If resetting the pipelineSlot steps isn't desirable then a check for existing steps might be needed before adding a step. If steps need to be manually removed then it might be helpful to have `removeStep`/`clearSteps`  methods.
[check existing / removing](https://github.com/aurelia/router/compare/master...z-brooks:pipeline)

The best place to remove steps might be the root `deactivate` method, but that doesn't get called when switching. Perhaps adding something in the `setRoot` method around https://github.com/aurelia/framework/blob/master/src/aurelia.js#L118 to call the viewModel deactivate if it exists?

EisenbergEffect:
Perhaps we can just have a reset method on the provider to set it back to the default configuration?

z-brooks:
That's what I'm doing now. Either way is fine, just rambling ideas. As long as there's a way to clear out unneeded/duplicate steps. Duplicates could be prevented with a simple check though.

EisenbergEffect:
Sounds like a nice PR opportunity. Any interest in submitting an improvement to the pipeline provider?

z-brooks:
[...]

PR #334, opened 2016-04-26
feat(PipelineProvider): add methods for removing pipeline steps

z-brooks:
https://github.com/aurelia/router/issues/332

### DIFF
--- a/src/pipeline-provider.js
+++ b/src/pipeline-provider.js
@@ -10,6 +10,21 @@ import {
   ActivateNextStep
 } from './activation';
 
+
+class PipelineSlot {
+  steps = [];
+
+  constructor(container, name, alias) {
+    this.container = container;
+    this.slotName = name;
+    this.slotAlias = alias;
+  }
+
+  getSteps() {
+    return this.steps.map(x => this.container.get(x));
+  }
+}
+
 /**
 * Class responsible for creating the navigation pipeline.
 */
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
@bryanrsmith Can you review this PR?

src/pipeline-provider.js
```diff
@@ -43,34 +58,55 @@ export class PipelineProvider {
...
-      static slotAlias = alias;
-      static steps = [];
-
-      constructor(container) {
-        this.container = container;
-      }
+  /**
+   * Removes a step from a slot in the pipeline
+   */
+  removeStep( name:string, step: PipelineStep ) {
```
bryanrsmith: If you could match the existing style by using a single space after colons, and no space inside parens that would be great.

src/pipeline-provider.js
```diff
@@ -43,34 +58,55 @@ export class PipelineProvider {
...
+    }
+  }
 
-      getSteps() {
-        return PipelineSlot.steps.map(x => this.container.get(x));
-      }
+  /**
+   * Clears all steps from a slot in the pipeline
+   */
+  clearSteps( name:string = '' ) {
```
bryanrsmith: Should this be private? It seems like a dangerous thing to allow.

bryanrsmith:
Left a couple comments, but otherwise looks good to me.

EisenbergEffect:
@z-brooks Could you address the comments that @bryanrsmith left? If so, I think we could get this in.

z-brooks:
Changes made.

EisenbergEffect:
Thanks! @jedd-ahyoung or @bryanrsmith Can you do a final review?

jedd-ahyoung:
@EisenbergEffect I checked this out.

@z-brooks This actually looks quite good, but I think that the class declaration here (https://github.com/aurelia/router/pull/334/files#diff-0b0b3e28fdaaf40a0b48ac8397c14380R14) needs to have the static properties defined in order for the factory (https://github.com/aurelia/router/blob/master/src%2Fpipeline-provider.js) to work. Placing the properties inside the constructor (i.e. the instance) means that they won't be available until the slots are instantiated. Have you verified that the tests are passing and that this is working with the skeleton-navigation app?

Additionally, we need to ensure that we have a `static inject` for the container instance.

Other than those issues, this looks good. (One of the problems here is that we have a lack of tests to catch this stuff.)

z-brooks:
@jedd-ahyoung All tests pass and it works fine in a skeleton. I'm also using this in my project (noted in the linked issue) without issue.

Not sure what you mean by a static inject for the container instance. `Container` is injected into the `PipelineProvider` [here](https://github.com/z-brooks/router/blob/pipeline-provider/src/pipeline-provider.js#L32) and I just pass it when constructing a new slot [here](https://github.com/z-brooks/router/blob/pipeline-provider/src/pipeline-provider.js#L110).

EisenbergEffect:
This is odd because when the provider loops through the steps it then tries to resolve an *existing instance* of PipelineStep from the container. However, it does work because when trying to resolve an object instance without a pre-registration the object instance will resolve to itself. So, it does work... We might want to put some comments to this affect here.

Issue #328, opened 2016-04-17
Observables in Router Activation Lifecycle


Ixonal:
I was thinking, it would be nice to have behavior similar to that of Promises in the activate hook, but for Observables (https://github.com/zenparsing/es-observable), with the whole not showing until the first value comes through. 

I figure it'd be something similar to this...
```javascript
let result = context.activate();
if(result && result.then instanceof Function) {
  //existing promise stuff
} else if(result && result.subscribe instanceof Function) {
  result.subscribe({
    next: (val, sub) => {
      sub.unsubscribe();
      completeActivate();
    }
  }); 
} else {
  //existing synchronous value stuff here
}
```

Admittedly, the unsubscribe part would require a modification to the spec to allow the subscription to be undone in the case of a synchronous first value. (https://github.com/zenparsing/es-observable/issues/83). Any thoughts?

EisenbergEffect:
We would be interested in a PR to add this if you find it useful.

Ixonal:
I'll look into it in a few days.

Ixonal:
Alright, found a way to handle the synchronous first value in subscriptions and made a fork using that https://github.com/Ixonal/router/blob/master/src/activation.js#L205

Any comments or concerns, or does it look good? If it's good, I'll set up the tests and put in a pull request.

EisenbergEffect:
That seems pretty reasonable. @bryanrsmith What do you think?

bryanrsmith:
Sounds reasonable to me, too. I haven't spent much time with observables, so I do have a few questions:

1. Is waiting for the *first* value the right thing to do, or will anyone want to wait for the sequence to complete?
1. Do we need to unsubscribe on error?
1. Do we need to observe sequence completion to cover the case where an observable produces no value?

Ixonal:
1. Considering that the Observable may never actually complete, waiting for that could be problematic. My thinking on this feature was in the way of using Observables with AJAX requests (like how Netflix and Angular 2 use them to get request cancellation support) and WebSockets. If it's for an AJAX request, the first value is the whole response and then it'd be complete anyways. If it's for a WebSocket, it'll most likely not complete for the duration of the page. In that case, the assumption is that the first message through means it's working and there should be enough information to render.

[...]

PR #336, opened 2016-05-14
feat(router): observables in lifecycle

Ixonal:
If an object matching an Observable's signature (according to the ECMA proposal) is returned from a custom element's activate hook, the router will wait until at least one piece of data has passed through the Observable to render the component. #328


Ixonal:
Ahh crap, hold on. Didn't notice the CI stuff earlier, fixing that.

PR #337, opened 2016-05-14
feat(router): observables in lifecycle

Ixonal:
feat(router): observables in lifecycle

When an Observabe (or something approximating one, as per the ECMA proposal) is returned from the activate hook, the router will wait until at least one value has moved through the Observable (or if the Observable completes while empty) before rendering the view.

Implements #336

### DIFF
--- a/src/activation.js
+++ b/src/activation.js
@@ -189,7 +189,7 @@ class SafeSubscription {
     this._subscribed = true;
     this._subscription = subscriptionFunc(this);
 
-    if (!this._subscribed) this.unsubscribe()
+    if (!this._subscribed) this.unsubscribe();
   }
 
   get subscribed() {
--- a/src/activation.js
+++ b/src/activation.js
@@ -180,9 +180,39 @@ function shouldContinue(output, router: Router) {
   return output;
 }
 
+class SafeSubscription {
+  //if this were TypeScript, subscriptionFunc would be of type 
+  //(sub: SafeSubscription) => Subscription
+  constructor(subscriptionFunc) {
+    this._subscribed = true;
+    this._subscription = subscriptionFunc(this);
+    
+    if(!this._subscribed) this.unsubscribe()
+  }
+  
+  unsubscribe() {
+    if(this._subscription) this._subscription.unsubscribe();
+    
+    this._subscribed = false;
+  }
+}
+
 function processPotential(obj, resolve, reject) {
   if (obj && typeof obj.then === 'function') {
     return Promise.resolve(obj).then(resolve).catch(reject);
+  }
+  
+  if(obj && typeof obj.subscribe === 'function') {
+    //an object with a subscribe function should be assumed to be an observable
+    new SafeSubscription(sub => obj.subscribe({
+      next() {
+        sub.unsubscribe();
+        resolve(obj);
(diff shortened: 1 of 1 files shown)


Ixonal:
the Travis-CI check is failing due to an existing issue in the test suite. Would've looked at fixing it, but that's sorta out of the scope of this pull request...

EisenbergEffect:
@bryanrsmith Can you do a final review on this?

EisenbergEffect:
ping @bryanrsmith for a final review.

src/activation.js
```diff
@@ -180,11 +180,58 @@ function shouldContinue(output, router: Router) {
   return output;
 }
 
+//wraps a subscription, allowing unsubscribe calls even if
+//the first value comes synchronously
+class SafeSubscription {
+  //if this were TypeScript, subscriptionFunc would be of type
+  //(sub: SafeSubscription) => Subscription
```
bryanrsmith: We use flow type annotations. If it's useful information it'd be great to add flow types and remove this comment.

src/activation.js
```diff
@@ -180,11 +180,58 @@ function shouldContinue(output, router: Router) {
...
 function processPotential(obj, resolve, reject) {
   if (obj && typeof obj.then === 'function') {
     return Promise.resolve(obj).then(resolve).catch(reject);
   }
 
+  if (obj && typeof obj.subscribe === 'function') {
+    //an object with a subscribe function should be assumed to be an observable
+    return new SafeSubscription(sub => obj.subscribe({
+      next() {
+        if (sub.subscribed) {
```
bryanrsmith: The check seems unnecessary since either state is handled in `unsubscribe`

src/activation.js
Ixonal: I haven't used Flow typings explicitly. Would it be the same type declaration as what TypeScript uses?

bryanrsmith: Yeah, they're pretty similar. Gulp will spit out a d.ts file that you can review if you want to confirm.

src/activation.js
Ixonal: you mean the check for whether the subscription is active? It's really there to avoid an extra resolve or reject (say, for a single value then complete), just kinda put the unsubscribe call in the block because... well, because.

src/activation.js
Ixonal: Mmmmkay, I'll have to look into it tomorrow, I'm headdin out for tonight.

src/activation.js
bryanrsmith: Ah, nevermind. I thought we were resolving a promise, but it's part of the pipeline. You're probably right that it's better to avoid resolving multiple times.

src/activation.js
Ixonal: Alright, added in type information. As far as I can tell, it looks about right in the definition file.

Ixonal: @bryanrsmith pinging just in case.

bryanrsmith:
Looks good to me

EisenbergEffect:
@Ixonal Thanks for this great contribution!

Ixonal:
@EisenbergEffect Not a problem :D

Issue #246, opened 2015-11-11
Repeat.for doesn't rebind when switching routes in a child router


devteamexpress:
This is based on the latest aurelia version as of 11-10

Here is my child viewmodel

    export class ChildView {
    
    constructor() {
    }

    configureRouter(config, router) {
        config.map([           
            { route: ['','route1'], name: 'route1', moduleId: 'views/configuration/route1/index', nav: true, auth: true, title: 'General'},            
            { route: ['route2','route2'], name: 'route2', moduleId: 'views/configuration/route2/index', nav: true, title: 'Communication'},            
        ]);
        this.router = router;
    }

    activate() {
                    
    }

}

Here is my view

    <template>

    <ul>
        <li repeat.for="item of router.navigation"><a href.bind="item.href">${item.title}</a></li>
    </ul>

    <router-view></router-view>
  
  </section>
     </template>

When i first load the page it works perfectly, if i switch to route2, the repeat.for doesn't rebind leaving me with no links to navigate back and forth.

The same thing happens with any repeat.for , i created a simple string array and this didn't either rebind.

bryanrsmith:
@devteamexpress Are you sure you're up to date on all the Aurelia libraries? That routing setup is exactly what we use in the skeleton navigation project, which appears to be working. If you are up to date, would you be able to post a plunker that reproduces the problem? I'm not able to repro myself.

xiwc:
i have the same question.

parent router change, the child router will not refresh bind.

my config:

      "aurelia-binding": "github:aurelia/binding@^0.10.0",
      "aurelia-bootstrapper": "github:aurelia/bootstrapper@^0.18.0",
      "aurelia-dependency-injection": "github:aurelia/dependency-injection@^0.11.0",
      "aurelia-event-aggregator": "github:aurelia/event-aggregator@^0.9.0",
      "aurelia-framework": "github:aurelia/framework@^0.17.0",
      "aurelia-history": "github:aurelia/history@^0.8.0",
      "aurelia-history-browser": "github:aurelia/history-browser@^0.9.0",
      "aurelia-i18next": "github:aurelia/i18n@^0.2.6",
      "aurelia-loader": "github:aurelia/loader@^0.10.0",
      "aurelia-loader-default": "github:aurelia/loader-default@^0.11.0",
      "aurelia-logging": "github:aurelia/logging@^0.8.0",
      "aurelia-metadata": "github:aurelia/metadata@^0.9.0",
[...]

PR #346, opened 2016-05-26
fix(router): change refresh navigation to check config for static href not nav model href

simonfox:
`router.refreshNavigation` should update the href of navigation model when route changes. Static routes (defined in config) do not need to update this property. The check for a static href was looking at the nav href instead of the config href.

closes aurelia/router#246

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -339,7 +339,7 @@ export class Router {
 
     for (let i = 0, length = nav.length; i < length; i++) {
       let current = nav[i];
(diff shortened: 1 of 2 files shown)


simonfox:
the CI failure is due to an unrelated test that was already failing...

EisenbergEffect:
Thanks @simonfox !!! It's going out tomorrow.

simonfox:
Awesome thanks @EisenbergEffect!!

Issue #326, opened 2016-04-14
Type Definition issue : Cancelling navigation


maverix:
Hi All,

I have implemented a PipelineStep in order to intercept navigation and cancel the navigation in certain situations.  It is working well, except I had to edit the typescript definition for aurelia-router.d.ts.

In the current version (1.0.0-beta.1.2.1) the definition for Next expects a value to be passed into the cancel method.  i.e. 

  export interface Next {
    
    /**
      * Indicates the successful completion of the entire pipeline.
      */
    complete(result: any): Promise<any>;
    
    /**
      * Indicates that the pipeline should cancel processing.
      */
    cancel(result: any): Promise<any>;
    
    /**
      * Indicates that pipeline processing has failed and should be stopped.
      */
    reject(result: any): Promise<any>;
    
    /**
      * Indicates the successful completion of the pipeline step.
      */
    (): Promise<any>;
  }

In order for typescript to be happy I changed the Cancel method to 
cancel(result?: any): Promise<any>;

I'm not exactly sure how you are generating your typescript definitions, otherwise I would make a PR to fix this.

Cheers.

EisenbergEffect:
You can make a PR directly to the source code to add the typing info there.

maverix:
Ok, I got around to trying to make a pull request.  But I couldn't for the life of me workout where in the api.json to indicate that a parameter was optional.  Any pointers?

EisenbergEffect:
No need to edit the api.json. That is generated from the code itself. You just need to indicate that the param is optional in the code using a `?`, for example `method(param?: string)`

maverix:
@EisenbergEffect does this look right?

EisenbergEffect:
Closing since the fix was merged :)

PR #362, opened 2016-06-28
Flagged parameters as optional

maverix:
https://github.com/aurelia/router/issues/326


EisenbergEffect:
For the future, we'd love it if you could follow our commit message guidelines :) You can read about that int he contributors guide linked at the top of any PR of issue submission form here on Github.

Issue #375, opened 2016-07-22
Add TypeScript ViewModel Interface


EisenbergEffect:
@TylerBrinkley commented on [Fri Jul 22 2016](https://github.com/aurelia/framework/issues/481)

It would be nice if a TypeScript ViewModel interface was added for better visibility of available features. It might look something like the below. There's probably more methods that could be specified but I just haven't discovered them yet which would just show the utility of adding this.

    export interface ViewModel {
        activate?(): void;
        canActivate?(): boolean|Promise<boolean>;
        deactivate?(): void;
        canDeactivate?(): boolean|Promise<boolean>;
    }

---

@EisenbergEffect commented on [Fri Jul 22 2016](https://github.com/aurelia/framework/issues/481#issuecomment-234561320)

This is specifically related to the Router, so I'm moving it there. I'm not entirely sure the benefit since it wouldn't affect the runtime at all and it would force a deeper dependence in your code on our framework by making you implement the interface. I'd favor someone just creating it in their own project if they want.

TylerBrinkley:
Thanks Rob, obviously all methods defined in the interface would be optional which means users wouldn't need to implement the interface explicitly but it might help them to better discover what features are available to them. Are there more than just these 4 methods that a ViewModel can implement?

AshleyGrant:
Over in NG2 world, they do stuff like this. I've been thinking it wouldn't hurt to have this, as long as the framework doesn't require a user to implement this interface and it just serves as a reminder.

Here's the docs form ng2: https://angular.io/docs/ts/latest/guide/lifecycle-hooks.html their interfaces are optional, and I actually prefer the strategy @TylerBrinkley is suggesting.

Maybe the name of the interface could be different `Routable` or `RoutableComponent` or something?

EisenbergEffect:
`RoutableComponent` I think would be the name. If someone wants to submit the PR to router, that would be great. Perhaps a similar interface `Component` should be added to templating. I'd accept a PR for that as well.

StrahilKazlachev:
@EisenbergEffect Do you consider:
- `determineActivationStrategy()` part of `RoutableComponent` ?
- `RoutableComponent` extends `Component` ?

EisenbergEffect:
No way to extend from Component because that is part of templating and router has no hard dependency on that. So, they should be independent.

You can include `determineActivationStrategy()` as well as `getViewStrategy()` as part of `RoutableComponent`

atsu85:
[...]

PR #376, opened 2016-07-22
feat(interfaces.js): add RoutableComponent

StrahilKazlachev:
Added `RoutableComponent` interface to `interfaces.js`(wasn't sure where it should go) and `NavigationCommand` interface to `navigation-commands.js` - see #375.
Also added `activationStrategy` to `RouteConfig`.

### DIFF
--- a/src/interfaces.js
+++ b/src/interfaces.js
@@ -74,7 +74,7 @@ interface RouteConfig {
   caseSensitive?: boolean;
 
   /**
-  * Add to specify an activation strategy if the strategy is always the same and you don't want that
+  * Add to specify an activation strategy if it is always the same and you do not want that
   * to be in your view-model code. Available values are 'replace' and 'invoke-lifecycle'.
   */
   activationStrategy?: string;
@@ -82,37 +82,55 @@ interface RouteConfig {
   [x: string]: any;
 }
 
-interface RoutableComponent {
+type ShouldContinue = boolean|Promise<boolean>|PromiseLike<boolean>|NavigationCommand;
+type Potential = Promise<any>|PromiseLike<any>|IObservable;
+
+interface CanActivateRoutableComponent {
   /**
   * Implement this hook if you want to control whether or not your view-model can be navigated to.
   * Return a boolean value, a promise for a boolean value, or a navigation command.
   */
-  canActivate?: (params: any, routeConfig: RouteConfig, navigationInstruction: NavigationInstruction) => boolean|Promise<boolean>|PromiseLike<boolean>|NavigationCommand;
+  canActivate: (params: any, routeConfig: RouteConfig, navigationInstruction: NavigationInstruction) => ShouldContinue;
+}
 
+interface ActivateRoutableComponent {
   /**
   * Implement this hook if you want to perform custom logic just before your view-model is displayed.
   * You can optionally return a promise to tell the router to wait to bind and attach the view until
   * after you finish your work.
   */
-  activate?: (params: any, routeConfig: RouteConfig, navigationInstruction: NavigationInstruction) => Promise<any>|PromiseLike<any>|IObservable|void;
+  activate: (params: any, routeConfig: RouteConfig, navigationInstruction: NavigationInstruction) => Potential|void;
+}
 
+interface CanDeactivateRoutableComponent {
   /**
   * Implement this hook if you want to control whether or not the router can navigate away from your
   * view-model when moving to a new route. Return a boolean value, a promise for a boolean value,
   * or a navigation command.
   */
-  canDeactivate?: () => boolean|Promise<boolean>|PromiseLike<boolean>|NavigationCommand;
+  canDeactivate: () => ShouldContinue;
+}
 
+interface DectivateRoutableComponent {
   /**
   * Implement this hook if you want to perform custom logic when your view-model is being
   * navigated away from. You can optionally return a promise to tell the router to wait until
   * after you finish your work.
   */
-  deactivate?: () => Promise<any>|PromiseLike<any>|IObservable|void;
+  deactivate: () => Potential|void;
+}
 
+interface DetermineActivationStrategyRoutableComponent {
   /**
-  * Implement this hook if you'd like to give hints to the router about the activation when reusing
+  * Implement this hook if you want to give hints to the router about the activation strategy, when reusing
   * a view model for different routes. Available values are 'replace' and 'invoke-lifecycle'.
   */
-  determineActivationStrategy?: (params: any, routeConfig: RouteConfig, navigationInstruction: NavigationInstruction) => string;
+  determineActivationStrategy: (params: any, routeConfig: RouteConfig, navigationInstruction: NavigationInstruction) => string;
+}
+
+interface ConfigureRouter {
+  /**
+  * Implement this hook if you want to add sub-routes to your view.
+  */
+  configureRouter(config: RouterConfiguration, router: Router);
 }
--- a/src/interfaces.js
+++ b/src/interfaces.js
@@ -87,28 +87,28 @@ interface RoutableComponent {
   * Implement this hook if you want to control whether or not your view-model can be navigated to.
   * Return a boolean value, a promise for a boolean value, or a navigation command.
   */
-  canActivate?: (params: any, routeConfig: RouteConfig, navigationInstruction: NavigationInstruction) => boolean|Promise<boolean>|NavigationCommand;
+  canActivate?: (params: any, routeConfig: RouteConfig, navigationInstruction: NavigationInstruction) => boolean|Promise<boolean>|PromiseLike<boolean>|NavigationCommand;
 
   /**
   * Implement this hook if you want to perform custom logic just before your view-model is displayed.
   * You can optionally return a promise to tell the router to wait to bind and attach the view until
   * after you finish your work.
   */
-  activate?: (params: any, routeConfig: RouteConfig, navigationInstruction: NavigationInstruction) => Promise<any>|IObservable|void;
+  activate?: (params: any, routeConfig: RouteConfig, navigationInstruction: NavigationInstruction) => Promise<any>|PromiseLike<any>|IObservable|void;
 
   /**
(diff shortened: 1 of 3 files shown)


atsu85:
@StrahilKazlachev, take a look at my solution, where 
> I've taken slightly different approach. I created one interface for each lifecycle method (instead of single interface for all optional methods) so that when i add the interface to the class, then TypeScript compiler actually checks the correctness of the method signature as well.

 I've created two files
* [ViewHooks.d.ts](https://gist.github.com/atsu85/2777454d6ee215a1b9d3911ce157828d#file-viewhooks-d-ts-L1) contains interfaces for Screen Activation Lifecycle hooks that Views can use.
* [HtmlBehaviorHooks.d.ts](https://gist.github.com/atsu85/5d6e5d6dbde996aaa33c1538a3eed9cd#file-htmlbehaviorhooks-d-ts-L1) contains interfaces for lifecycle hooks that all HTML Behaviors (such as Views and Custom Elements) can use.

src/interfaces.js
```diff
@@ -73,5 +73,46 @@ interface RouteConfig {
...
+  
   [x: string]: any;
 }
+
+interface RoutableComponent {
+  /**
+  * Implement this hook if you want to control whether or not your view-model can be navigated to.
+  * Return a boolean value, a promise for a boolean value, or a navigation command.
+  */
+  canActivate?: (params: any, routeConfig: RouteConfig, navigationInstruction: NavigationInstruction) => boolean|Promise<boolean>|PromiseLike<boolean>|NavigationCommand;
```
atsu85: Looks like babel 6 also [supports type aliases](https://github.com/codemix/babel-plugin-typecheck/pull/43), like TypeScript, so the return type of canActivate and canDeactivate could be defined in one place and used for both methods, smth like I've done [here](https://gist.github.com/atsu85/2777454d6ee215a1b9d3911ce157828d#file-viewhooks-d-ts-L16)

src/interfaces.js
```diff
@@ -73,5 +73,46 @@ interface RouteConfig {
...
+  * navigated away from. You can optionally return a promise to tell the router to wait until
+  * after you finish your work.
+  */
+  deactivate?: () => Promise<any>|PromiseLike<any>|IObservable|void;
+
+  /**
+  * Implement this hook if you'd like to give hints to the router about the activation when reusing
+  * a view model for different routes. Available values are 'replace' and 'invoke-lifecycle'.
+  */
+  determineActivationStrategy?: (params: any, routeConfig: RouteConfig, navigationInstruction: NavigationInstruction) => string;
```
atsu85: Would the build (with Babel) still produce valid output If You'd replace the result type `string` with `'replace' | 'invoke-lifecycle'` to narrow down the return type to the minimal correct options?

StrahilKazlachev: @atsu85 No, tried it. I get an error from babel-dts-generator.

atsu85: Yeah, based on what i [found](https://github.com/codemix/babel-plugin-typecheck/pull/43), I would have also placed a bet against my hopes - @EisenbergEffect it is another reason to use TypeScript instead of Babel ;)

src/interfaces.js
```diff
@@ -132,5 +132,5 @@ interface ConfigureRouter {
   /**
   * Implement this hook if you want to add sub-routes to your view.
   */
-  configureRouter(config: RouterConfiguration, router: Router);
+  configureRouter(config: RouterConfiguration, router: Router): any;
```
StrahilKazlachev: I'm not sure how to type the return type. From what I saw it can be: `RouterConfiguration`, any, void, and Promise and PromiseLike that resolve to these.

atsu85: Then why don't You mark it with union type? As I understand union types are supported by Babel 6 (and TypeScript).

StrahilKazlachev: I don't see what's the benefit of doing union with `any`? The compiler will never throw an error. It's as good as a doc comment. Or am I missing something?

atsu85: > It's as good as a doc comment.

Yeah, except in code, where it belongs to ;) Also, as You see, You haven't specified it in the doc either.

StrahilKazlachev: So something like this should do?
```
type RouterConfigurationResult = RouterConfiguration|any;
interface ConfigureRouter {
  configureRouter(config: RouterConfiguration, router: Router): RouterConfigurationResult|Promise<RouterConfigurationResult>|PromiseLike<RouterConfigurationResult>;
}
``` 
I may be overthinking it but it doesn't need to return exactly `RouterConfiguration`, it just checks if the returned value has an `exportToRouter(router: Router): void` and calls it. I probably should just leave it as is.

atsu85: @StrahilKazlachev, I'm wondering, how did You come up with the return type of configureRouter method?
(By the way, it is the only method in [my sample](https://gist.github.com/atsu85/2777454d6ee215a1b9d3911ce157828d#file-viewhooks-d-ts-L19-L24) that isn't written by myself, but my teammate). 
The [official documentation](http://aurelia.io/hub.html#/doc/article/aurelia/framework/latest/cheat-sheet/6) has one place where the signature is: 
`configureRouter(config: RouterConfiguration, router: Router): void`
and five places where the second argument is omitted from the signature:
`configureRouter(config: RouterConfiguration): void`
but none of them has anything else but `void` as the return type.

From the source i found 2 places, that use that method:
* [route-loading.js#L109](https://github.com/aurelia/router/blob/34119d1ce827f10fe5d3c7c97133d83ad922eab6/src/route-loading.js#L109)
* [app-router.js#L66](https://github.com/aurelia/router/blob/34119d1ce827f10fe5d3c7c97133d83ad922eab6/src/app-router.js#L66)
but i didn't spend any time to figure out what are the potential return types. 

@EisenbergEffect, @bryanrsmith - could You clarify the return type of `configureRouter`?

StrahilKazlachev: [router.js#L117](https://github.com/aurelia/router/blob/master/src/router.js#L117)
Like I said, I'm probably digging too deep and `void` probably is good enough.
Also in your sample the `configureRouter` is missing explicit return type.

atsu85: > Also in your sample the configureRouter is missing explicit return type. 

Yes, i also discovered that myself, when writing my previous comment - as I said,
>By the way, it is the only method in my sample that isn't written by myself, but my teammate.

so it slipped through my code review.

But I'd still like to get a reply from the Aurelia team:
> @EisenbergEffect, @bryanrsmith - could You clarify the return type of configureRouter?

EisenbergEffect: `configureRouter` can actually return void or Promise.

src/interfaces.js
```diff
@@ -73,5 +73,65 @@ interface RouteConfig {
...
+  * Add to specify an activation strategy if it is always the same and you do not want that
+  * to be in your view-model code. Available values are 'replace' and 'invoke-lifecycle'.
+  */
+  activationStrategy?: string;
+  
   [x: string]: any;
 }
+
+type ShouldContinue = boolean|Promise<boolean>|PromiseLike<boolean>|NavigationCommand;
+type Potential = Promise<void>|PromiseLike<void>|IObservable;
```
atsu85: I guess it makes sense to add type info (`Potential`) to `processPotential` function as well - [activation.js#L222](https://github.com/aurelia/router/blob/e7d01b7fd73adf90feb2dc2301cdf372a8209b48/src/activation.js#L222)

StrahilKazlachev: Not much for me, but I'll add it for completeness and future references. But as `Potential|any` since the 2 `if`s exhaust `Potential` and in ts 2.0, after the last `if` the type will be `never`. I'll wait for any other input and if non any, I'll squash the commits or I shouldn't?

atsu85: > I'll wait for any other input

@EisenbergEffect, could someone from the Aurelia team review this PR as well? I don't have any more suggestions :)

> I'll squash the commits or I shouldn't?

Yeah, I guess You'll be asked to do so

EisenbergEffect:
What's the status of this?

atsu85:
I'd say it's good, but it could use another pair of eyes to rewiew

StrahilKazlachev:
@EisenbergEffect If you think it's ok and there are no problems with the interfaces naming I'll squash the commits.

EisenbergEffect:
I can squash on merge no problem. Let me review it one more time.
@niieani Do you have any time to review these TS convenience interfaces?

atsu85:
Oh yeah, the naming could also need attention

niieani:
I believe the interfaces should be exported, i.e. `export interface` not just `interface`. Skimming through them I'd say they're ok. In any case, they're only interfaces and we can make them better in the future if we need to. I'd say good to merge @EisenbergEffect.

StrahilKazlachev:
I did not use `export` since non of the other interfaces in this module use.

niieani:
Interesting. I guess the `babel-dts-generator` we use exports interfaces automatically? Could @cmichaelgraham confirm? TypeScript needs interfaces to be explicitly exported.

cmichaelgraham:
@niieani i think that is right - babel-dts-generator probably shouldn't automatically export, but it would mean big code changes at this point, so maybe we stay with the current behavior? @EisenbergEffect

StrahilKazlachev:
Just one quick question, since I'm preparing a PR with the `Component`(bind, attached, ...) interfaces and I prefer naming consistency. 
- current - router: `ActivateRoutableComponent`, ...; templating: `BindComponent`, `AttachedComponent`, ...
- router: `RoutableComponentActivate`, ...; templating: `ComponentBind`, `ComponentAttached`, ...

niieani:
Actually I think the second version, with the prefixing instead of suffixing is better.

StrahilKazlachev:
It turns out that type aliases are not generated in the definition file. If you try to `export` them you get `TypeAlias not supported` from the gulp build.

EisenbergEffect:
@StrahilKazlachev What do you think about calling it `ConfiguresRouter`?

Issue #366, opened 2016-07-06
Generate route should use route variables


jods4:
I have the following route definition:
`{ route: 'location/:id', name: 'location' }`

But when I call `router.navigateToRoute('location', { id: 42 })` it navigates to: `/location?id=42`.

This does work for Aurelia, but it would be better if it went to the expected route: `/location/42`.
Other components (and users) might look at the URL and this is a bit unexpected.

EisenbergEffect:
It should work as you say, so there must be some sort of bug.

jods4:
Found the bug. It was in our code.

Someone created different routes for new and existing entities with the same name:
```javascript
config.map([
  { router: 'location/:id', name: 'location', ... },
  { router: 'location', name: 'location', ... }
]);
```

It seems the last one wins, which then makes perfect sense.
I renamed the second route with `name: 'newLocation'` and now it works as it should, indeed.

Should there be at least a warning when you use the same name for different routes?

EisenbergEffect:
Probably. We hadn't considered that sort of mistake :) We can certainly add it. Can you create a new issue for that? Also, this sounds like a great opportunity for a community PR :)

jods4:
@EisenbergEffect we can re-open this issue if you want to track the warning issue?

jods4:
#377 has been merged and is published in 1.0.0 but it contains a critical bug!

Apparently, the patch #377 trips on single routes that have multiple alternative URLs.
So trying to register this route:
`{ route: ['', 'home'], name: 'home' }`
Triggers the distinct name error.

**This is a very hard regression, because the impact is that the application does not start at all.**

It is also a bit tricky to figure out, because the error message is clear, but doesn't say what route name is faulty... and when you look at your code, there is no duplicate names.

/cc @devanp92

EisenbergEffect:
It's already fixed. We realized it during the release process and reverted that code entirely.

EisenbergEffect:
Update to 1.0.2

Alexander-Taran:
can be closed.. maybe

davismj:
This is still an issue and a good one to track.

PR #377, opened 2016-07-24
fix(router): ensure routes have distinct names

devanp92:
This is in reference to #366.

Router will throw an error if a route with the same name in addRoute

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -226,7 +226,7 @@ export class Router {
   * @param navModel The [[NavModel]] to use for the route. May be omitted for single-pattern routes.
   */
   addRoute(config: RouteConfig, navModel?: NavModel): void {
-    validateRouteConfig(config);
+    validateRouteConfig(config, this.routes);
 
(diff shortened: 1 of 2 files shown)


CasiOo:
Should the check be a part of `function validateRouteConfig()`?
Should the comparison be case insensitive? Different casing is not tested.

devanp92:
@CasiOo 

> Should the check be a part of function validateRouteConfig()?

That is a good point regarding design. I can make that change if necessary.

> Should the comparison be case insensitive? Different casing is not tested.

Interesting point. @EisenbergEffect can you chime in?

EisenbergEffect:
If possible, it's probably better to locate that inside the existing validation method. As far as case sensitivity goes, the recognizer's name lookup is currently case-sensitive.

devanp92:
@EisenbergEffect Thank you. I have updated the code per your comment.

In validateRouteConfig(), I also needed to pass Router's routes since it was not contained in the validateRouteConfig's context.

EisenbergEffect:
Thank you!

PR #381, opened 2016-07-29
make router have the responsibility of resolving the location of the …

kristianmandrup:
Works in conjunction with https://github.com/aurelia/templating-router/pull/35

Allows Router to provide strategy for how to find a viewModel, no longer being limited/hardcoded to the path relative to the router file.

A use case would be f.ex to have reusable components installed in `src/resources/components`.
It quickly becomes ugly and tiresome to prepend with the full path.

`{ route: 'contacts',  moduleId: 'resources/components/contact-detail', name:'contacts' }`

Better then to subclass the `Router` with a different. `viewModelLocation` strategy, perhaps using a custom attribute on the route to determine how to find the route module.

`{ route: 'contacts',  moduleId: 'contact-detail',  name:'contacts', component: true }`

Would also be preferable to avoid having all the components in a flat structure. How about a folder per viewModel as an alternative. Again a custom router strategy could allow for this while keeping the route configuration clean.

```bash
components/
  contact-detail/
     contact-detail.html
     contact-detail.ts
  contact-list/
     ...
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
       "aurelia-history": "^1.0.0",
       "aurelia-logging": "^1.0.0",
       "aurelia-path": "^1.0.0",
+      "aurelia-metadata": "^1.0.0",
       "aurelia-route-recognizer": "^1.0.0"
     },
     "dependencies": {
--- a/src/router.js
+++ b/src/router.js
@@ -9,6 +9,8 @@ import {
   _createRootedPath,
   _resolveUrl} from './util';
 import {RouteConfig} from './interfaces';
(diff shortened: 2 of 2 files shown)


kristianmandrup:
This change will make it much easier to work with components, see [aurelia-installer](https://www.npmjs.com/package/aurelia-installer)

EisenbergEffect:
@kristianmandrup You had a couple of changes related to this. Is this one also obsolete? Can you point me to the latest proposal?

kristianmandrup:
This one: https://github.com/aurelia/templating-router/pull/38

PR #384, opened 2016-08-03
RouteConfig interface doesn't contain auth property

arnederuwe:
the router :: RouteConfig interface doesn't contain the auth property, therefore giving an error in TypeScript. Added a cast to any to correct this, as you probably want to have a dedicated property for this stuff instead of adding it to the settings object, which would pollute things a bit.

### DIFF
--- a/doc/article/en-US/router-configuration.md
+++ b/doc/article/en-US/router-configuration.md
@@ -422,7 +422,7 @@ A pipeline step must be an object that contains a `run(navigationInstruction, ne
     }
 
     class AuthorizeStep {
-      run(navigationInstruction: NavigationInstruction, next: Function): Promise<any> {
+      run(navigationInstruction: NavigationInstruction, next: Next): Promise<any> {
         if (navigationInstruction.getAllInstructions().some(i => (<any>i.config).auth)) {
           var isLoggedIn = //insert magic here;
           if (!isLoggedIn) {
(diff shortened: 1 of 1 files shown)


andrewtobin:
I think it would be better adding the auth property to the interface as it is referenced for other versions of the code (ESnext).  I've created a new PR as I just hit the same problem: https://github.com/aurelia/router/pull/385

And yes, I should have probably checked someone hadn't already addressed it - my bad :)

arnederuwe:
I'm not sure if adding an auth property by default is a good idea
I don't think it should be part of this interface as auth doesn't really have to do anything with routing, it's more an extension when you are using authentication.
The core reason why I wouldn't include it though is that a boolean property won't satisfy everyone's needs, say for example that you are using role based authorization. Then you would probably like a 'Role' string property or even a custom object instead of a boolean to provide specific users access to specific routes.
I also changed the Function to Next, forgot to edit it here :)

andrewtobin:
Good argument, I didn't even think past the sample in some ways.  If I'm wrong, I'm wrong - I'm just getting started in both TypeScript and Aurelia and the buggy samples are interesting to get around.

EisenbergEffect:
The presence of an `auth` property directly on the route config in the docs is an error. That property should be on the `settings` object which is the designated location for custom properties.

Issue #349, opened 2016-06-06
Customizing the Navigation Pipeline in Typescript: "Property 'cancel' does not exist on type 'Function'."


vaags:
Hi. I'm trying to customize the Navigation pipeline in my main router, as described in the docs:(http://aurelia.io/docs.html#/aurelia/framework/1.0.0-beta.1.2.4/doc/article/securing-your-app).

I'm using TypeScript with the latest [typescript-webpack skeleton](https://github.com/aurelia/skeleton-navigation/tree/master/skeleton-typescript-webpack) (1.0.0-beta.1.2.6).

I've entered `config.addPipelineStep('authorize', AuthorizeStep);` to add an authorization step inside the configureRouter function.

The step (AuthorizeStep) is being called just as expected. But, inside AuthorizeStep, when I call `return next.cancel(new Redirect('login'));`, I get a squiggly red line underneath "cancel" (in VS Code), and the following error:

`[ts] Property 'cancel' does not exist on type 'Function'.`

Any ideas? This might just be a rookie mistake, but I really just pasted in the code from the docs - which should be valid.

EisenbergEffect:
The code is correct, but our d.ts files might not be, which is causing TS to complain.

vaags:
Ah. I see. Could this be related to #326? I tried changing the aurelia-router.d.ts, like @maverix did there, but still got the squiggly.

EisenbergEffect:
Typings should be all fixed up now.

andrewtobin:
@EisenbergEffect I've just installed the cli, and tried this and still getting the same problem.  I've made sure to do an npm update on the aurelia-router but same.

andrewtobin:
It looks like the sample code on the site should read more like:
  class AuthorizeStep {
    run(navigationInstruction: NavigationInstruction, next: Next): Promise<any> {

But I _think_ that auth needs to be added onto the RouteConfig interface? Not sure - I'm pretty new to TypeScript. https://github.com/aurelia/router/blob/master/dist/aurelia-router.d.ts#L87

EisenbergEffect:
We need to update the samples there. Any custom settings, like auth, should be added onto the `settings` property.

Issue #265, opened 2015-12-23
AuthorizeStep shows TSC errors like Property 'auth' does not exist on type 'RouteConfig'.


HIRANO-Satoshi:
There is the following code in the cheat sheet (typescript). I have similar code borrowed from the cheat sheet.

http://aurelia.io/docs.html#/aurelia/framework/1.0.0-beta.1.0.6/doc/article/cheat-sheet

```
class AuthorizeStep {
  run(navigationInstruction: NavigationInstruction, next: Function): Promise<any> {
    if (navigationInstruction.getAllInstructions().some(i => i.config.auth)) {
      var isLoggedIn = /* insert magic here */false;
      if (!isLoggedIn) {
        return next.cancel(new Redirect('login'));
      }
    }

    return next();
  }
}
```

tsc shows the following errors.

    Property 'auth' does not exist on type 'RouteConfig'.
    Property 'cancel' does not exist on type 'Function'.

How do I fix?

bryanrsmith:
Looks like we need to adjust the type annotations on the next parameter. It's a function with a `cancel` property tacked on, but it's declared as a function right now.

HIRANO-Satoshi:
I see.

bryanrsmith:
@cmichaelgraham I believe that [this line](https://github.com/aurelia/router/blob/master/src/pipeline.js#L24) needs to use a type annotation like

```typescript
run(instruction: NavigationInstruction, next: { (): Promise, cancel: (result: any) => Promise, reject: (result: any) => Promise, complete: (result: any) => Promise }): void;
```

but the dts generator doesn't appear to support it. Is this something we can add easily, or is there alternative syntax that might work?

bryanrsmith:
Thanks, @cmichaelgraham. Looks like your change worked great.

HIRANO-Satoshi:
Thank you very much!

PR #385, opened 2016-08-05
Adjust documentation and interface for TypeScript Authentication

andrewtobin:
fix(TypeScript interface): Added auth property to match code sample.

The documentation refers to a method signature showing "next" described
as "Function" when an interface was introduced to wrap it and make sure
it could be cancelled.

There's also an update to the interfaces to place the "auth" property on
the RouteConfig so that the code sample works.

Fixes issues: #349 #265   -- they're closed, but I got this bug in most recent version.

### DIFF
--- a/doc/article/en-US/router-configuration.md
+++ b/doc/article/en-US/router-configuration.md
@@ -422,7 +422,7 @@ A pipeline step must be an object that contains a `run(navigationInstruction, ne
     }
 
     class AuthorizeStep {
-      run(navigationInstruction: NavigationInstruction, next: Function): Promise<any> {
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
I can accept the doc change but not the addition of the auth property. That somehow slipped into the docs and should not be there. It should be on the `settings` property.

EisenbergEffect:
Actually, there shouldn't be an auth property on the route config. That should be on settings. We need to fix up the docs if it's showing that. I can take this PR if you can revert the interfaces change and just have the doc fix. Sound good?

EisenbergEffect:
Closing this with no response from author. Docs have been corrected.

Issue #386, opened 2016-08-10
Missing viewPorts when using navigationStrategy and child router


StrahilKazlachev:
**I'm submitting a bug report**

* **Library Version:**
1.0.2

**Please tell us about your environment:**
* **Operating System:**
Windows [7|10]

* **Node Version:**
6.2.0

* **NPM Version:**
3.8.9

* **JSPM OR Webpack AND Version**
JSPM 0.16.41

* **Browser:**
all

* **Language:**
all

**Current behavior:**
- using `navigationStrategy` to dynamically determine a `moduleId`
- dynamically provided VM has child router
- if navigated to one of the child routes:
 - on app load(first load) - the application just hangs, no errors, no rejected promises
 - after successfully navigating to any route an error is thrown [here](https://github.com/aurelia/router/blob/37ab9fadc16c884335e4531bb0c934621926d795/src/navigation-plan.js#L43) - `config.viewPorts` is `undefined`.

Here is a [gist](https://gist.run/?id=95102d5df0bd4a80318f7168818f8d0d). When you go to **Dynamic Root** select **Dynamic B II** and you can see the error in the console.

**Expected/desired behavior:**
The child routes to load, like if a `moduleId` and not `navigationStrategy` was provided

StrahilKazlachev:
After some digging it turns that my problem comes from [here: JSON.parse(JSON.stringify(config));](https://github.com/aurelia/router/blob/37ab9fadc16c884335e4531bb0c934621926d795/src/router.js#L254). This strips the `navigationStrategy` and later when a `NavigationInstruction` is being build it does not get its `viewPorts` set. Changing it to `Object.assign` resolves the problem. Is there a reason for stripping away functions or is it something done before `navigationStrategy` was added?

StrahilKazlachev:
@EisenbergEffect Could someone comment the issue. We are still stuck on this.

Issue #289, opened 2016-02-09
Incorrect routing when using navigationStrategy callback


WouterOuwens:
Currently, I'm looking into dynamic routing in aurelia. To achieve this, I mapped all routes using a wildcard and I'm using the navigationStrategy property to determine the correct module id (as suggested in https://github.com/aurelia/router/issues/129). 

``` javascript
configureRouter(config: RouterConfiguration, router: Router): void {
    config.map([
        { route: "**", nav: false, navigationStrategy: this.navigateToModule, settings: this }
    ]);

    router.handleUnknownRoutes(this.unknownRoutes);
}

private navigateToModule(instruction: NavigationInstruction): void {

// some logic to determine a moduleId
instruction.config.moduleId = moduleId; // A valid view-model
}
```

The first navigation is correct and loads the correct viewmodel, as expected. However, when I now navigate to _another_ route in the same Router (with another moduleId), it always navigates to the _current_  active viewmodel (instead of the _new_ one). The navigationStrategy callback is called witch each navigation, as expected. 

The activate() function of the current (incorrect) viewmodel is called, while the navigationInstruction.config.moduleId in the activate() function contains the _new_ moduleId (as set in the navigateToModule function). When refreshing the site, the correct viewmodel is loaded. 

The code is running on the current Aurelia release (beta.1.1.0), using Typescript. I'm encountering this in all browsers. Is this expected behavior?

I also tried to map the routes using the handleUnknownRoutes (as sugested in https://github.com/aurelia/router/issues/15). This is working as expected. However, I prefer to use the described way, since it is possible to use to settings property in the router better (there is no reference to the original class in the handleUnknownRoutes callback) and it is possible to use multiple wildcards and redirect them to other functions.

bryanrsmith:
@WouterOuwens Are you able to provide a plunker or link to a repo that demonstrates the problem?

EisenbergEffect:
Closing this issue since it was not reproducible and the OP didn't respond after several months.

Roustalski:
@WouterOuwens How did you get around this issue? I am running up against the same problem. I'll see if I can create a reproducible scenario, forked from the TS skeleton, for the team.

Roustalski:
OK, I have a reproducible scenario: https://github.com/Roustalski/skeleton-navigation

`cd path/to/skeleton-navigation-typescript
npm install && jspm install
gulp watch`

Click on child router, then modify the route URL:
http://localhost:9001/#/child-router/1
[...]

PR #393, opened 2016-08-30
Updates the navigation strategy to handle and expect a resolved strin…

Roustalski:
…g or object of viewport names with module ids for each key.

Primarily for [#289](https://github.com/aurelia/router/issues/289), but also [#386](https://github.com/aurelia/router/issues/386)


Roustalski:
I was going to also write some unit tests for this, but I couldn't get the tests to run locally after doing a jspm install. Not sure what that's about.

EisenbergEffect:
Is there some reason why the custom navigation strategy can't just add the view ports onto the instruction's config? That was the original intent.

StrahilKazlachev:
When I started using the `navigationStrategy` I wondered why it didn't work like `mapUnknownRoutes`, which I find really flexible. Even with this change if you want to set the title you have to modify the instruction NavModel.
@EisenbergEffect You say that the original intent was to add the view ports but I can't find an example in the hub. The only example I found with `navigationStrategy` does
```
instruction.config.moduleId = instruction.fragment
instruction.config.href = instruction.fragment
```
Which does not work if you to return different `moduleId` on each call(depending on the parameter/s). If you say that augmenting the instruction is the way to go I can add some more to the docs.

EisenbergEffect:
Can you foresee any issues with just adding a viewPorts property onto the config in the callback?

Roustalski:
From the app developer's perspective, all they want is for a module to be loaded based on logic within their callback. At least this is the only use-case I've needed so far.

Why not set the appropriate places on the instruction for them based on the resolved string or viewport-ID:module-id object mapping? It's just sugar.

http://aurelia.io/hub.html#/doc/article/aurelia/router/latest/router-configuration/3
```
    var navStrat = (instruction) => {
      instruction.config.moduleId = instruction.fragment
      instruction.config.href = instruction.fragment
    }
```
Changes to:
```
    var navStrat = (instruction) => {
      return Promise.resolve().then(() => instruction.fragment));
    }
```
If the `viewPorts` property is added, then it changes to:
```
    var navStrat = (instruction) => {
      instruction.config.viewPorts['default'] = instruction.fragment;
    }
```

EisenbergEffect:
@Roustalski Are you able to sign the CLA? There's a link in the details section of this issue.

Roustalski:
Alright, I signed the CLA by following the details link and clicking the button, but it still says it hasn't been signed O.o

EisenbergEffect:
@Roustalski Is there any chance the email associated with your local git isn't the same as Github? We've had this happen a couple of times so we are trying to track down the source. @PWKad Can you chime in on this based on your experience?

Roustalski:
That was the problem. I have rebased the branch with amended commits with the reset-author option after updating my config to my github name and email

EisenbergEffect:
@bryanrsmith See any issues with this?

src/router.js
```diff
@@ -452,13 +452,21 @@ function validateRouteConfig(config: RouteConfig, routes: Array<Object>): void {
...
-        'default': {
-          moduleId: instruction.config.moduleId
-        }
-      };
+      instruction.config.viewPorts = {};
+    }
+
+    if (typeof modules === 'string') {
+      modules = {'default': modules};
+    } else if (modules === undefined) {
```
davismj: For example, if someone has logic in their nav strat that falls back to a null value, they might expect it to run with the default config, but instead, modules will remain null and this will throw an unhandled error.

I'd recommend updating this to check for a falsey value.

EisenbergEffect: @Roustalski Can you update to handle this case?

Roustalski:
Yea, I'll check this out and update.

davismj:
@Roustalski Hey Russ, appreciate the time you took on this PR. Did you have a chance to update this PR?

sorenhoyer:
@StrahilKazlachev I made [this repo](https://github.com/sorenhoyer/aurelia-dynamic-routing) to replicate another problem I currently have, but I added a console.log(instruction) in app.js on line 24, which - as I understand it - would indicate the exact opposite of what [you state here](https://github.com/aurelia/router/issues/289#issuecomment-238684345): _"The first time navigationStrategy is called navigationInstruction.config does not have viewPorts set"_. 

If you download the repo and run it, you should see that it works just fine - or at least as I would expect it to work. 

Maybe I'm just missing your point entirely? :-)

StrahilKazlachev:
@sorenhoyer There were quite a lot of fixes since my comment.

Alexander-Taran:
Stale?

EisenbergEffect:
Defer to @davismj to determine what to do with this.

davismj:
Since the PR was never updated I can't merge it in. However, I don't have my head in this issue well enough to know if it is important or not. I'd like to come back to it later, but without more interest its low prio.

Issue #297, opened 2016-02-24
Wildcard routes not working?


Souldrinker:
I'm trying out to switch my login logic to an external identity server to whom I pass in a redirect url back to my application which means I get a call like ```http://localhost:4703/#id_token=atoken&scope=openid%20profile&state=somestate&session_state=somesessionstate```.

In this case I wanted to catch it with a wild card route like ```{ route: "id_token*token", moduleId: "views/login/oauth-check/oauth-check", nav: false, title: "Validation" }```
...in which I expected to get in the rest of the url after ```id_token=``` passed in as <param>.token to my activate method in the oauth-check class from which I could strip out my token and use it for further calls to my web api. But no matter what I try I only get route not found errors.

The workaround I did for now is to configure mapUnknownRoutes  in which I check the fragment for id_token and in that case returns my module id. In that way I get the whole path as a property on the indata parameter in my activate method which is fine for me. But shouldn't wildcard routes have worked like I described above?

bryanrsmith:
Wildcard segments can be used to match the "rest" of the path, but they only work with complete segments. I.e., `foo/*path` would match `/foo/bar/baz`, with the `path` parameter holding `'bar/baz'`. partial segment matchign with wildcards, like `foo/bar*path`, is currently not supported.

Also note that routing only inspects the path portion, not the query string. The URL in your example looks like it's storing a query string in the hash (though the `?` is missing).  Are you able to change the URL structure and use something like `/auth?id_token=atoken&scope=...`, and then match the route on `auth`? If not, then your `mapUnknownRoutes` solution may be the best option.

Souldrinker:
Ok, thanks for the quick response. Then I'll stick with my mapUnknownRoutes that works fine with a few lines of code.

The reason I asked was that I had Google authentication working in one of my old Durandal projects with the following setup and thought it worked the same in Aurelia:
```
import router = require("plugins/router");

class Shell {
    activate(): any {
        var routes = new Array<DurandalRouteConfiguration>();
        routes.push({ route: "", moduleId: "viewmodels/login", nav:false });
        routes.push({ route: "login", moduleId: "viewmodels/login", nav: false });
        routes.push({
            route: "selfRegistration(/:email/:firstName/:lastName)",
            moduleId: "viewmodels/SelfRegistration/SelfRegistration", nav: false
        });
        routes.push({ route: "access_token=*token", moduleId: "viewmodels/OauthCheck/OauthCheck", title: "Validation", nav: false });
[...]

PR #395, opened 2016-09-01
doc(router): fix wildcard router config and documentation

Thanood:
Hi all,

I just tried to make wildcard routes to work, had a look at the documentation and saw that this route won't work:
```javascript
{ route: 'files*path',       name: 'files',      moduleId: 'files/index',   href:'#files',   nav: true },
```

I've searched around and found this comment from @bryanrsmith:
https://github.com/aurelia/router/issues/297#issuecomment-188447021

I derived from that comment that the configuration mentioned above should rather look like this:
```javascript
{ route: 'files/*path',       name: 'files',      moduleId: 'files/index',   href:'#files',   nav: true },
```

I've tried that in my app and it works.

This PR is my attempt to fix the router documentation.
If you'd rather like to adjust the router implementation to match the documentation, please feel free to close this PR. :smile:

### DIFF
--- a/doc/article/en-US/router-configuration.md
+++ b/doc/article/en-US/router-configuration.md
@@ -592,7 +592,7 @@ A pipeline step must be an object that contains a `run(navigationInstruction, ne
       configureRouter(config, router) {
         config.title = 'Aurelia';
         config.map([
-          { route: 'users', name: 'users', viewports: { left: { moduleId: 'user/list' }, right: { moduleId: 'user/detail' } } }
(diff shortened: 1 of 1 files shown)


atsu85:
I also ended up acting the same way as @Thanood - glad someone already sent the PR :)

EisenbergEffect:
@Thanood Good catch! We definitely need to get this fix into the docs. Would you mind updating to the master branch? It seems there are some conflicts with other fixes.

Thanood:
Sure, will do that.

PR #402, opened 2016-09-15
fix(typings): RoutableComponentCanActivate.canActivate should accept …

atsu85:
…`Promise<NavigationCommand>` as return type


StrahilKazlachev:
Shouldn't this change be done in `src/interfaces.js`

atsu85:
@StrahilKazlachev, absolutely, and that is where i thought i did the changes :D Thanks for pointing it out, now it is fixed.

*As an excuse, I can only say, that after my initial change in interfaces.js file, my IDE reformatted whole file, so I had to revert it. After turning off auto-formatter, I must have navigated to the code by interface name, that only came up from dist folder - dooh!*

PR #406, opened 2016-09-20
Publishes an inner instruction resolve

Roustalski:
I have a route that simply redirects to a different module, but I wasn't receiving any published events from the router with the instruction details.

This allows me to hook in on the instruction that is being re-directed to without having to worry about why there is a conditional on the other events that get published.

### DIFF
--- a/src/app-router.js
+++ b/src/app-router.js
@@ -199,9 +199,9 @@ function processResult(instruction, result, instructionCount, router) {
 function resolveInstruction(instruction, result, isInnerInstruction, router) {
   instruction.resolve(result);
 
+  let eventArgs = { instruction, result };
   if (!isInnerInstruction) {
     router.isNavigating = false;
-    let eventArgs = { instruction, result };
     let eventName;
 
     if (result.output instanceof Error) {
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
@bryanrsmith Can I bug you to review this? I'd appreciate your feedback.

bryanrsmith:
> I have a route that simply redirects to a different module, but I wasn't receiving any published events from the router with the instruction

Perhaps we should instead consider a code change that would ensure an event was published for redirects, rather than always publish events for child navigations. It could either be a custom event (`router:redirect:complete`?), or the standard event(s) with the redirect instruction. It might be a clearer API to consume than the more technical child navigation events.

I don't object to publishing events for child instructions if that's desirable, though. If we stick with this I think the event name could use a small adjustment, since it uses some terminology that refers to implementation detail. Maybe `router:navigation:innerInstruction:resolve` -> `router:navigation:child:complete`, or `router:childNavigation:complete`?

EisenbergEffect:
@bryanrsmith What if we did both? Add a new redirect event and then also had child events? I'm usually in favor of enriching the information that the system is providing about what it's doing. Thoughts?

bryanrsmith:
Sounds good to me. 👍

EisenbergEffect:
@Roustalski Any interest in working on this further?

Roustalski:
I updated the event name per the suggestion, but I was unable to figure out where to place another event for `router:redirect:complete`. It seems that the redirect instruction simply gets canceled and then the pipeline moves on and it wasn't clear to me when the redirect event should fire.

EisenbergEffect:
@bryanrsmith Any thoughts on how to handle the complete event in this case?

PR #409, opened 2016-09-22
feat(router): add extension to transform title when built

### LINKED ISSUE #343: Localizing title
This is somewhat related to #218. 

Currently, I'm using some kind of resource identifier the page title in the router config:
`
config.map([
            {
                route: ["", "home"],
                name: "home",
                moduleId: "dashboard/home",
                nav: true,
                title: "home"
            },
`
I'm localizing the title in the nav-bar.html using t.bind, and this works. 

However, this does not work with the title of the website, which Aurelia builds up automatically from the NavigationInstructions using the title field from the config.

Is there any way to localize the page title?

manuel-guilbault:
Add a `transformTitle: (title: string) => string` property to the `Router`. Can be used to connect the router's title generation with a i18n library.

Example:

``` js
import Backend from 'i18next-xhr-backend';
import {AppRouter} from 'aurelia-router';

export function configure(aurelia) {
  aurelia.use
    .standardConfiguration()
    .plugin('aurelia-i18n', i18n => {
      i18n.i18next.use(Backend);

      return i18n.setup({
        backend: {
          loadPath: './locales/{{lng}}/{{ns}}.json', 
        },
        lng : 'en',
        fallbackLng : 'en',
        attributes : ['t','i18n'],
        debug : false
      }).then(() => {
        aurelia.container.get(AppRouter).transformTitle = title => i18n.tr(title);
      });
    })
    .feature('resources');

  aurelia.start().then(() => aurelia.setRoot());
}
```

Resolves #343


EisenbergEffect:
@manuel-guilbault Excellent contribution! We'll do review soon and get some feedback to you. Thanks!

src/router.js
```diff
@@ -333,6 +335,16 @@ export class Router {
...
+  _transformTitle(title: string) {
+    if (this.transformTitle) {
+      return this.transformTitle(title);
+    }
+    if (this.parent) {
+      return parent.transformTitle(title);
+    }
+    return title;
+  }
+
```
stsje: We could probably remove _transformTitle, and instead let transformTitle have the default implementation before/if it is overridden. That way we won't have any confusion if internal objects should be using _transformTitle or transformTitle.

src/navigation-instruction.js
```diff
@@ -228,16 +228,20 @@ export class NavigationInstruction {
   }
 
   _buildTitle(separator: string = ' | '): string {
-    let title = this.config.navModel.title || '';
+    let title = '';
     let childTitles = [];
 
+    if (this.config.navModel.title) {
```
stsje: What if the title is the String '0', will it evaluate to false, or does it work as expected?

manuel-guilbault: The string '0' will be evaluated to true, like any unempty string will.

src/router.js
manuel-guilbault: Good point. I'll change that. Thx :)

dist/amd/aurelia-router.js
```diff
@@ -377,16 +377,20 @@ define(['exports', 'aurelia-logging', 'aurelia-route-recognizer', 'aurelia-depen
     NavigationInstruction.prototype._buildTitle = function _buildTitle() {
       var separator = arguments.length <= 0 || arguments[0] === undefined ? ' | ' : arguments[0];
 
-      var title = this.config.navModel.title || '';
```
atsu85: don't commit files in dist folder

manuel-guilbault: @atsu85 sorry, didn't know that. Fixed.

src/router.js
```diff
@@ -333,6 +335,16 @@ export class Router {
     return undefined;
   }
 
+  _transformTitle(title: string) {
+    if (this.transformTitle) {
+      return this.transformTitle(title);
+    }
+    if (this.parent) {
+      return parent.transformTitle(title);
```
EisenbergEffect: Should this line be a call to `this.parent._transformTitle(title)`?

manuel-guilbault: I changed that in reaction to some comments. Instead of leaving a ```transformTitle``` property and a ```_transformTitle``` method trying to use this property, I removed the ```_transformTitle``` method and put its implementation as the default value of the ```transformTitle``` property. I feel like it's clearer, and it's lesser lines of code.
Do you see any downsides I missed?

manuel-guilbault:
Any issue with this PR? If not, any idea when it will be merged and released?

EisenbergEffect:
@manuel-guilbault Can you add something to this PR that updates the documentation on how to use this feature?

EisenbergEffect:
I want to be careful about improving our docs whenever we add new features. That's something that often gets forgotten and then no body knows that the feature exists. If you can add a section to the docs on this, then we can probably get it merged.

manuel-guilbault:
Sure, I completely agree. I'll update the PR as soon as I can update the doc. Thx!

manuel-guilbault:
@EisenbergEffect done :)
I don't know how I can "test" the additonal doc. I tried to mimic existing sections.

timfish:
Some of the code looks like it's ended up outside the code blocks:

image

src/navigation-instruction.js
```diff
@@ -233,16 +233,20 @@ export class NavigationInstruction {
...
+    }
+
     for (let viewPortName in this.viewPortInstructions) {
       let viewPortInstruction = this.viewPortInstructions[viewPortName];
 
       if (viewPortInstruction.childNavigationInstruction) {
         let childTitle = viewPortInstruction.childNavigationInstruction._buildTitle(separator);
         if (childTitle) {
-          childTitles.push(childTitle);
+          childTitles.push(this.router.transformTitle(childTitle));
```
EisenbergEffect: It seems like this would cause a double translation of the child title wouldn't it? The first would happen inside the _buildTitle, since it calls transformTitle internally, and then the second would happen on line 249. Is that correct or am I missing something? @manuel-guilbault

manuel-guilbault: @EisenbergEffect right, good catch. The code has changed since I wrote this, I didn't see this side effect when I updated my fork. I just committed the fix for this. Thx!

PR #414, opened 2016-10-02
Update router-configuration.md

lares83:
There is a mistake in Layout section.
Instead of 
```<router-view layout="views/layout-default.html"></router-view>```
there should be
```<router-view layout-view="views/layout-default.html"></router-view>```


lares83:
Closing as duplicate of #396

PR #415, opened 2016-10-03
fix broken example code

dissolve:
Example code doesn't work this fixes
https://github.com/aurelia/documentation/issues/279


EisenbergEffect:
Thanks for helping!

PR #416, opened 2016-10-05
Update router-configuration.js

lpj145:
#134 line have an expression error.


EisenbergEffect:
`exportToRouter` is the correct method name.

PR #417, opened 2016-10-05
Update router-configuration.md

lpj145:
Its works best!


EisenbergEffect:
Not quite correct since we want to check the truthiness of the value.

PR #421, opened 2016-10-08
Syntax error when using the example provided in docs

leosbrf:
fix router/doc/article/en-US/router-configuration.md: unexpected token when building the module


EisenbergEffect:
Thanks all!

PR #422, opened 2016-10-09
doc(interfaces): fix optional interfaces

Gheoan:


atsu85:
No, no, no! This is absolutely against the original idea of these interfaces: if You add the interface to the class, then you MUST IMPLEMENT the method (with correct signature)! That is why we have many interfaces (with required methods) instead of one (with optional methods) in first place!

Gheoan:
I don't understand why would anyone want multiple interfaces.

My use case is:
```ts
// someotherfile.ts
import {
  RoutableComponentCanActivate,
  RoutableComponentActivate,
  RoutableComponentDeactivate,
  RoutableComponentCanDeactivate,
  RoutableComponentDetermineActivationStrategy
} from 'aurelia-router';

export interface RoutableComponent extends RoutableComponentCanActivate,
  RoutableComponentActivate,
  RoutableComponentDeactivate,
  RoutableComponentCanDeactivate,
  RoutableComponentDetermineActivationStrategy { }

// my-component.ts
import {RoutableComponent} from './someotherfile';

class MyClass implements RoutableComponent {
  ...
}
```

With multiple interfaces you have to import all of them for each component and be explicit:
```ts
// my-component.ts
import {
  RoutableComponentCanActivate,
  RoutableComponentActivate,
  RoutableComponentDeactivate,
  RoutableComponentCanDeactivate,
  RoutableComponentDetermineActivationStrategy
} from 'aurelia-router';

class MyClass implements RoutableComponentCanActivate,
  RoutableComponentActivate,
  RoutableComponentDeactivate,
  RoutableComponentCanDeactivate,
  RoutableComponentDetermineActivationStrategy {
  ...
}
```

StrahilKazlachev:
Because if a member is optional and you implement it wrong the compiler will not complain. It accepts that it is **not** implemented - since optional, not that it is **incorrectly** implemented. I prefer the overhead of a bit more code than loosing the type safety.

Gheoan:
If a method with the same name exists both in the interface and in the class that implements the interface,  and if the method has wrong parameter types or wrong return type, even when the method is optional, typescript will still warn that the interface is not corectly implemented. Otherwise it is a bug with typescript.

The only way to lose type information is having a typo in the method name.

Alternatively, an interface could be added that extends these interfaces and make the methods optional only there.

EisenbergEffect:
This is pretty contentious so I'm going to leave things as they are for now.

PR #427, opened 2016-10-24
doc(typescript): fix mapUnknownRoutes Typescript compile error

ErikSchierboom:
The TypeScript examples that shows how to setup a function to map unknown routes has an error in it. At the moment, it returns an object containing a `moduleId` string. However, this leads to a compile error, which makes sense if we examine the typings:

``` typescript
mapUnknownRoutes(config: string | RouteConfig | ((instruction: NavigationInstruction) => string | RouteConfig | Promise<string | RouteConfig>)): RouterConfiguration;
```

The specific function overload used is expected to return a `string`, not an object. The ES6 example returns the same object, so either the typings are wrong or there is also a bug in the ES6 example.

### DIFF
--- a/doc/article/en-US/router-configuration.md
+++ b/doc/article/en-US/router-configuration.md
@@ -301,8 +301,7 @@ The function passed to `mapUnknownRoutes()` has to return:
         config.title = 'Aurelia';
 
         var handleUnknownRoutes = (instruction) => {
-            // return 'not-found';
-            return { moduleId: 'not-found' };
+            return { route: 'not-found', moduleId: 'not-found' };
         }
 
         config.mapUnknownRoutes(handleUnknownRoutes);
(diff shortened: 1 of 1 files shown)


doc/article/en-US/router-configuration.md
```diff
@@ -326,9 +326,8 @@ The function passed to `mapUnknownRoutes()` has to return:
 
         config.title = 'Aurelia';
 
-        let handleUnknownRoutes = (instruction: NavigationInstruction): {moduleId: string} => {
```
davismj: This isn't wrong, it is just a (necessary?) simplification of the following.

```let handleUnknownRoutes = (instruction: NavigationInstruction): RouteConfig => {```

Though your fix here is also not wrong, it doesn't match the ES2015 version of the code. I would update both places to match, or just fix the typing for the TypeScript version.

StrahilKazlachev: 1(why hide a correct solution) and 3(`route` is like the **id** of the `RouteConfig`) are no go for me. Adding a route field is more acceptable for me. There is a lot that can be improved in the typings, but it is hard with the current generator.

ErikSchierboom: I've updated the PR to use option number 2, which seems like a sensible solution.

Issue #431, opened 2016-11-10
Fallback route in scenario without previous location set


MT-PL-SWF-Dev:
Our application uses deep linking. During activate() step there is possibility that promise can be rejected and in this situation normally Router will return to previous location.
However if there is no previous location (e.g. when F5 is pressed), application omits any other navigation going straight to url location. In this scenario we recieve "Router navigation failed, and no previous location could be restored.".

We would love to have possiblity to configure fallback route for this scenario to handle such scenarios. Ideally it could be configured per child-router also so that if location is nested inside one it won't go to generic default route but specific for this part of application.

With regards,
Maciej

MT-PL-SWF-Dev:
I tried to do this on live code and I have managed to make it work with quite simple solution. Maybe it would give some light on my idea or maybe even suggest way for applying it.

**RouterConfiguration**
`configFallbackRoute(route: string) {
      this.fallbackRoute = route;
      return this;
    }`

`exportToRouter(router: Router) {`
`...`
`      router.fallbackRoute = this.fallbackRoute;`
`...`
`}`

**AppRouter**
```
function restorePreviousLocation(router) {
    var previousLocation = router.history.previousLocation;
    if (previousLocation) {
      router.navigate(router.history.previousLocation, { trigger: false, replace: true });
    } else {
      if(router.fallbackRoute) {
        router.navigate(router.fallbackRoute, { trigger: true, replace: true });
      } else {
        logger.error('Router navigation failed, and no previous location could be restored.');
      }
    }
  }
```

EisenbergEffect:
Would you be interested in submitting a PR to add this feature?

MT-PL-SWF-Dev:
Of course if you agree on this solution I will create one :)

vuorinem:
This is really useful feature and works nicely, except when the fallback route is empty.

In https://github.com/aurelia/router/blob/69c3b4456e3150ed6bb2380458b994c4e34c46cb/src/app-router.js#L233, The line:
```
} else if (router.fallbackRoute) {
```
could be comething like this to allow empty routes:
```
} else if (router.fallbackRoute !== undefined && router.fallbackRoute !== null) {
```

shanonvl:
@vuorinem if empty routes must be supported, [this logic in RouteConfiguration](https://github.com/aurelia/router/blob/master/src/router-configuration.js#L159) will need to be changed as well.    

[...]

PR #432, opened 2016-11-14
Fallback route config

MT-PL-SWF-Dev:
Proposal of solution for enhancement:
https://github.com/aurelia/router/issues/431

### DIFF
--- a/doc/article/en-US/router-configuration.md
+++ b/doc/article/en-US/router-configuration.md
@@ -688,3 +688,44 @@ Similar to MVC-style master/layout pages, Aurelia allows configuration of multip
     }
   </source-code>
 </code-listing>
+
+## [Configuring Fallback Route](aurelia-doc://section/11/version/1.0.0)
+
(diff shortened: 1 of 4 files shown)


EisenbergEffect:
I'm happy to merge this but with one request. Before we add a new feature, I'd like to have documentation for it. In the docs folder you will find an article on router configuration. Can you add a section there that covers this feature and add that to this PR? If you can do that we can get it in. I just don't want it to go undocumented. Thanks!

MT-PL-SWF-Dev:
Sure, I will do this soon :)

MT-PL-SWF-Dev:
I cannot merge those conflicts, however I think that everything is ready.

stsje:
Wouldn't it be a good idea to implement the feature the same way as `mapUnknownRoutes` http://aurelia.io/hub.html#/doc/article/aurelia/router/latest/router-configuration/6 ?
So we'll have a `mapFallbackRoutes` function.

EisenbergEffect:
If you can submit a PR for that, I think that would be a better way to handle it, yes.

PR #470, opened 2017-02-14
chore(empty-routes): fix child urls with emptys

### LINKED ISSUE #27: Let child routers run off empty URL fragments
As part of implementing authentication, I was trying to have a login page that looked different to the authenticated part of the application. To do this I set up a default route that went to a layout page which then had a child router that contained the other routes; for example, 'dashboard', 'orders', 'returns' etc.

In my main router configuration I had a default route and a login route:

``` js
{ route: '', moduleId: './layout', nav: false },
{ route: 'login', moduleId: './login', nav: false }
```

Then in the `layout` module I had defined a child router with a 'dashboard' route:

``` js
{ route: 'dashboard', moduleId: './dashboard', nav: true, title: 'Dashboard' }
```

When trying to navigate to `#dashboard` I was receving the following error: `Route Not Found: dashboard`.

I think this is a fairly common scenario and may not even be used for authentication, somebody could have a landing page that looked completely different to the rest of their site and expect something like this to work.

Perhaps the framework could check the root router for the route first, then do the child checks taking into account an empty fragment for the default route and then go through the same logic it currently uses to find child routes now and then if no route is found, it throws a "Route not found" error.

PWKad:
Fixes #27

When a child router is checking for a parents baseUrl it does not take
in to account that the parent may be routing to the empty route.  This
PR adds a spec to cover that scenario and a basic implementation.

### DIFF
--- a/src/navigation-instruction.js
+++ b/src/navigation-instruction.js
@@ -161,19 +161,29 @@ export class NavigationInstruction {
   * Gets the instruction's base URL, accounting for wildcard route parameters.
   */
   getBaseUrl(): string {
+    let fragment = this.fragment;
+
+    if (fragment === '') {
+      let nonEmptyRoute = this.router.routes.find(route => {
+        return route.name === this.config.name &&
+          route.route !== '';
+      });
+      fragment = nonEmptyRoute.route;
+    }
+
     if (!this.params) {
-      return this.fragment;
+      return fragment;
     }
 
     let wildcardName = this.getWildCardName();
     let path = this.params[wildcardName] || '';
 
     if (!path) {
-      return this.fragment;
+      return fragment;
     }
 
     path = encodeURI(path);
-    return this.fragment.substr(0, this.fragment.lastIndexOf(path));
+    return fragment.substr(0, fragment.lastIndexOf(path));
   }
 
   _commitChanges(waitToSwap: boolean) {
--- a/src/navigation-instruction.js
+++ b/src/navigation-instruction.js
@@ -168,7 +168,9 @@ export class NavigationInstruction {
         return route.name === this.config.name &&
           route.route !== '';
       });
-      fragment = nonEmptyRoute.route;
+      if (nonEmptyRoute) {
+        fragment = nonEmptyRoute.route;
+      }
     }
 
     if (!this.params) {
--- a/test/navigation-instruction.spec.js
+++ b/test/navigation-instruction.spec.js
@@ -1,0 +1,95 @@
+import {History} from 'aurelia-history';
+import {Container} from 'aurelia-dependency-injection';
+import {AppRouter} from '../src/app-router';
+import {PipelineProvider} from '../src/pipeline-provider';
+import {NavigationInstruction} from '../src/navigation-instruction';
+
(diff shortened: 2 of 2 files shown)


src/navigation-instruction.js
```diff
@@ -161,19 +161,29 @@ export class NavigationInstruction {
...
   */
   getBaseUrl(): string {
+    let fragment = this.fragment;
+
+    if (fragment === '') {
+      let nonEmptyRoute = this.router.routes.find(route => {
+        return route.name === this.config.name &&
+          route.route !== '';
+      });
+      fragment = nonEmptyRoute.route;
```
EisenbergEffect: What happens if it doesn't find a route?

PWKad: Do you mean there is a situation that supports an empty route fragment or that we should handle in case one is not met - 

```
if (!nonEmptyRoute) {
  return throw new Error('Matching route not found!');
}
fragment = nonEmptyRoute.route;
```

PWKad:
Ok fixed that issue and re-enabled all tests.

smithaitufe:
I am this issue has been resolved.I have a common scenario here but I am still facing the same issue.

I am using aurelia-router version 1.2.1.

I don't know if I am doing something wrong.

app.ts
```
configureRouter(config: RouterConfiguration, router: Router) {
    config.title = 'Aurelia';
    config.map([
      { route: '', name: 'main',      moduleId: './main-app/app',      nav: true, title: 'Welcome' },
      { route: '/login', name: 'login',      moduleId: './login/login',      nav: true, title: 'Login' }
    ]);

    this.router = router;
  }
```
then I have 

main-app/app.ts

```
configureRouter(config: RouterConfiguration, router: Router) {
    config.title = 'Aurelia';
    config.map([
      { route: ['', 'welcome'], name: 'welcome',      moduleId: './welcome/welcome',      nav: true, title: 'Welcome' },
      { route: '/packages', name: 'packages',      moduleId: './packages/packages',      nav: true, title: 'Packages' }
    ]);

    this.router = router;
  }
```
 However, I can't go to packages.  The first route resolves properly. This means the route `welcome` is loaded properly when the app starts.
Then, I can go to `login` route. 

But the problem is I can't go to `packages` route. Rather it is displaying this in the console: 
```
ERROR [app-router] Error: Route not found: /packages
    at AppRouter._createNavigationInstruction (aurelia-router.js:1035)
    at AppRouter.loadUrl (aurelia-router.js:1630)
    at BrowserHistory._loadUrl (aurelia-history-browser.js:284)
    at BrowserHistory._checkUrl (aurelia-history-browser.js:277)
error @ aurelia-logging-console.js:54
(anonymous) @ aurelia-logging.js:42
(anonymous) @ aurelia-router.js:1633
tryCatcher @ bluebird.js:5253
Promise._settlePromiseFromHandler @ bluebird.js:3275
Promise._settlePromise @ bluebird.js:3332
Promise._settlePromise0 @ bluebird.js:3377
Promise._settlePromises @ bluebird.js:3452
(anonymous) @ bluebird.js:175
(anonymous) @ bluebird.js:4507
```

Your help is appreciated.

Issue #443, opened 2016-12-09
Global Unknown Route handler with child routers and partial route matches?


ShaunVanDyk:
**I'm submitting a bug report**

* **Library Version:**
aurelia-router 1.1.0

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
6.2.0

* **NPM Version:**
4.5.0

* **JSPM OR Webpack AND Version**
Neither

* **Browser:**
all

* **Language:**
TypeScript 2.2.0-dev.20161125

**Current behavior:**
I've setup a module to handle unknown routes and configured the root router in app.ts:
config.mapUnknownRoutes("./shared/shell/not-found");

I also register some children routes, and those children then setup their own routes, e.g.:
config.map([
  {
    route: "child-one",
    moduleId: "./app_features/child_one/child_one" // Hits the ts file and registers its child routes
  },
  {
    route: "child-two",
    moduleId: "./app_features/child_two/child_two" // Hits the ts file and registers its child routes
  }
]);

If I navigate to http://localhost:9000/non-existent-url or http://localhost:9000/non-existent-child/non-existent-url, then the unknown route works perfectly.
If I navigate to http://localhost:9000/child-one/non-existent-url, it logs an error saying the route isn't found but doesn't hit the not-found module.

If I call config.mapUnknownRoutes while setting up the child routes, then it will render the unknown module inside the router-view of that child.

Setting up a callback to get the Navigation Instruction (and do a redirect) has the same behaviour.

**Expected/desired behavior:**

I was hoping to register an unknown route page to redirect to regardless of whether the route partially matches a child route or not.

EisenbergEffect:
Essentially, it sounds like if a child router can't find the route and doesn't have a not found handler, you want it to "bubble" up the parent router chain looking for a not found handler. Is that correct?

ShaunVanDyk:
Yeah pretty much :)

Or more realistically a setting to opt-in for it to bubble up, so that it doesn't break for anyone using custom not found pages per child router (if that's a use case).

gheoan:
@EisenbergEffect, @davismj this can be closed.

PR #474, opened 2017-02-24
feat(router): enable parent chain unknown route handlers

jagonalez:
bubble up the current router parent chain for the first defined unknown route handler if one does not exists on the current router

feature #443


CasiOo:
Aurelia's `_commitChanges` method is becoming pretty complicated. Maybe the method should be refactored making it more readable and easier to understand?

EisenbergEffect:
@bryanrsmith Any chance to review this and let me know what you think?

matik12:
@bryanrsmith Any info on this enhancement?

EisenbergEffect:
@Vheissu @davismj Can you either of you review this?

PR #485, opened 2017-04-01
feat(viewport): add viewport name to life cycle arguments

balazsmeszegeto:
Consider the following scenario:

The same view-model can be reused in different view ports (for example in a two-column layout). The rendering might slightly differ depending on which view port is it being rendered in. How should the view-model access this information?

Added viewport's name to config object being sent to `activate` / `deactivate`


EisenbergEffect:
Thanks!

Issue #489, opened 2017-04-12
Incorrect window.history.state data in IE11


TomFoyster:
**I'm submitting a bug report**

**Library Version:**
"aurelia-router": "npm:aurelia-router@^1.0.0",

**Operating System:**
|Windows 7

**Node Version:**
6.10.1

**NPM Version:**
3.10.10

**JSPM OR Webpack AND Version**
JSPM 0.16.53

**Browser:**
Internet Explorer 11

**Language:**
ESNext

**Current behavior:**
- I am a user, and I store data in `window.history.state`.
- I then organically navigates to another page in the App, and then navigate back to the original page by using a link (not the browsers back button).
- I then view the data stored in `window.history.state`.
- In IE11 I see the data I stored in the earlier step

**Expected/desired behavior:**
As this page view should be a new entry in window.history I shouldn't see any data stored in `window.history.state`

I have tested this outside Aurelia - with 2 simple HTML pages using `replaceState()` and it performs as expected in IE11.

An example App to replicate this can be found; 
https://github.com/TomFoyster/AureliaHistoryStateTest

jwx:
Pretty sure this is based to different behaviours in the browsers. Looking into it.

EisenbergEffect:
I don't think this has anything to do with Aurelia.

jwx:
@EisenbergEffect Not sure I agree. Aurelia's router is responsible for navigation within the app and the problem occurs since it doesn't take into consideration that IE/Edge behaves differently when using location.hash to "create navigation". I do however think I could create a PR that fixes this, if you want it.

EisenbergEffect:
Please provide :)

TomFoyster:
I'd also argue this shouldn't be closed until a solution is implemented. 

If this were normal behaviour in IE11 outside of Aurelia I'd agree that it isn't an issue However, as covered by @jwx, the role of this package is to replicate the navigation behaviour one would usually encounter when using a non-SPA site. Part of that behaviour is `window.history.state` being cleared when a new page is loaded. 

Closing it because you _think_ it's nothing to do with Aurelia is a little premature IMO.

jwx:
@EisenbergEffect @TomFoyster It wasn't quite as straight forward as I had hoped, but there's now two PRs that will consolidate the behaviour of IE/Edge with that of Chrome and Firefox.

EisenbergEffect:
@jwx Thanks! We'll check them out and try to get them into the next set of releases.

[...]

PR #490, opened 2017-04-13
fix(router): browser history state not empty on new entries in IE/Edge

jwx:
The router doesn't handle the different behaviour in IE/Edge for history state on new entries. This fix consolidates the behaviour with Chrome etc.

Depending on PR aurelia/history-browser/history-state-consolidation.
Closes aurelia/router#489.


EisenbergEffect:
@davismj This is related to the other PR I pinged you on.

PR #491, opened 2017-04-17
enhancement(router): enable lazy load feature when navigating to route

jmvtrinidad:
Load module with id configure.ts/js in the base folder from route.
It will enable to lazy load features/plugin/element.

import `configurationLoaderStep` then in app.ts
```typescript
config.addPreActivateStep(new ConfigurationLoaderStep(this.aurelia));
```

example configure.ts
```typescript
import { FrameworkConfiguration } from 'aurelia-framework';

export function configure(config: FrameworkConfiguration) {
    return config        
        .globalResources('./ad-content/about')
        .apply();
}

```

Enhancement aurelia/framework#427


EisenbergEffect:
The idea here is great. One problem is that we can't have a dependency on the framework from inside the router. This is something that probably needs to be implemented inside of the framework, in a duck typed way. It would also be good to unify this with view resources somehow so that views could "require" configuration. Finally, we'd probably want the framework to apply the configuration, rather than have the developer do that, so it matches normal feature modules more. I'm closing this for now, but I'd love to see another PR that addresses the feedback above. I'm sure the community would love this.

jmvtrinidad:
Hi @EisenbergEffect 
Sorry, I don't have an idea on how it will unify with view resources and others thing maybe you can give some light on how to do it? Technically how Aurelia repo relates to each other seems a magic to me.
Thank you.

Issue #411, opened 2016-09-26
deactivate() does not invoked after navigate to other router module.


jdanyow:
@xiwc commented on [Fri Aug 19 2016](https://github.com/aurelia/framework/issues/563)

**I'm submitting a bug report**
- **Library Version:**
  "aurelia-framework": "^1.0.0-rc.1.0.0",

**Please tell us about your environment:**
- **Operating System:**
  Windows [7|10]
- **Node Version:**
  v4.4.2
- **NPM Version:**
  3.10.5
- **JSPM OR Webpack AND Version**
  webpack 1.13.1
- **Browser:**
  Chrome 52.0.2743.116 m
- **Language:**
  ESNext

**Current behavior:**
when i switch router.
1. `#/server/disk  -> #/server/other`  deactivate() is invoked normally
2. `#/server/disk -> #/network/other` deactivate() is not invoked normally

---

@guilhermewaess commented on [Tue Sep 13 2016](https://github.com/aurelia/framework/issues/563#issuecomment-246892236)

Are you using layouts?
I have the same problem, but with attached() <- attached() never calls when I using layouts :(

---

@alpox commented on [Wed Sep 14 2016](https://github.com/aurelia/framework/issues/563#issuecomment-246932154)

Seems to be a duplicate of https://github.com/aurelia/templating-router/issues/34

mbenedyk:
any progress here?

MT-PL-SWF-Dev:
I have made some research and found out that this bug is probably caused by if clauses in "activation.js" file. https://github.com/aurelia/router/blob/master/src/activation.js#L72

```javascript
if (viewPortPlan.childNavigationInstruction) {
      findDeactivatable(viewPortPlan.childNavigationInstruction.plan, callbackName, list);
    } else if (prevComponent) {
      addPreviousDeactivatable(prevComponent, callbackName, list);
}
```

If we navigate across AppRouter from nested Child Router _viewPortPlan.childNavigationInstuction_ is _undefined_ while running _canDeactivate()_ but before _deactivate()_ is called there is called _loadRoute()_ method which sets _viewPortPlan.childNavigationInstuction_
After this _deactivate()_ step kicks in but does not call _addPreviousDeactivatable(prevComponent, callbackName, list);_ but rather search for deactivatable in plan for _childNavigationInstruction_!

In our project it is quite serious scenario which cannot be resolved by decent workarounds.

[...]

Issue #132, opened 2015-06-22
Deactivate method isn't called when user is moving between two child routers.


milunka:
I think that deactivate method isn't called on page for this situation:
- application has 3 routers defined as one master router with two child routers
- user moves from page in one child router to different page in different child router and deactivate method on page where user was at the begining isn't called.

Example of how child routers are configured:

**app.js**:

```
config.map([
  { route: '',  moduleId: './welcome',      nav: true, title:'Welcome' },
    { route: 'child1', moduleId: './child1', nav: true, title: 'Child 1' },
    { route: 'child2', moduleId: './child2', nav: true, title: 'Child 2' },
  ]);
```

**child1.js**

```
config.map([
  { route: '', moduleId: './page1', nav: true, title:'Page 1' },
  { route: 'page1a', moduleId: './page1a', nav: true, title:'Page 1a' },
]);
```

**child2.js**

```
config.map([
  { route: '', moduleId: './page2', nav: true, title:'Page 2' },
]);
```

When user is on _Page 1a_ and moves to _Page 2_ this is list of called events:
- child1.js - deactivate
- child2.js - activate
- page2.js - activate
- child1.js - detached
- page1a.js - detached
- child2.js - attached
- page2.js - attached

I think that as first method must be called deactivate method on _Page 1a_ and then all methods which are called now.

Demo of this problem: https://github.com/milunka/redirect-activate-problem

brantwedel:
I am having this same problem in a slightly different scenario:

When a parent route changes with a replace strategy:
```js
determineActivationStrategy() { return activationStrategy.replace; }
```

deactivate does not get called on child VM's, but since `canDeactivate` *is* called in this scenario, I am using this workaround:

```
class VM {
    deactivate() {
        // deactivation logic, ensure this code does not have side effects if called twice
    }
    canDeactivate() {
        return this.deactivate() || true;
    }
}
```

derekpitt:
Is this still an issue? I am seeing the same thing here too..

heruan:
I confirm this, `deactivate` isn't called between VMs on the same navigation hierarchy.

tomtomau:
Also echoing the same issue as @heruan

heruan:
Will this possibly be addressed in the next release? Since now I was using @brantwedel trick, but I'm facing a scenario where I really need `deactivate` to be called *only if* `canDeactivate` returns true!

EisenbergEffect:
@bryanrsmith Can you look at this with top priority?

[...]

PR #496, opened 2017-05-03
fix(router): invoke deactivate method on child modules

jagonalez:
This PR adds additional logic to determine the current step and whether to prioritize prevComponent or childNavigationInstruction

The 'canDeactivate' step should prioritize child navigation instructions whereas the 'deactivate' step should prioritize the previous component which will contain the additional child modules.

I had originally swapped the statements around like this:
```
    if (prevComponent) {
      addPreviousDeactivatable(prevComponent, callbackName, list);
    } else if (viewPortPlan.childNavigationInstruction) {
      findDeactivatable(viewPortPlan.childNavigationInstruction.plan, callbackName, list);
    }
```
But this caused one of the canDeactivate tests to fail. 

This PR addresses issues: 
[#132](https://github.com/aurelia/router/issues/132) 
[#411](https://github.com/aurelia/router/issues/411)

### DIFF
--- a/src/activation.js
+++ b/src/activation.js
@@ -58,6 +58,7 @@ function findDeactivatable(plan, callbackName, list: Array<Object> = []): Array<
   for (let viewPortName in plan) {
     let viewPortPlan = plan[viewPortName];
     let prevComponent = viewPortPlan.prevComponent;
+    let canDeactivate = callbackName === "canDeactivate";
 
     if ((viewPortPlan.strategy === activationStrategy.invokeLifecycle ||
         viewPortPlan.strategy === activationStrategy.replace) &&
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
Sweet! I'll check this out. Thanks!

guidocecilio:
Is there a specific date for a release of the aurelia-router component that includes this fix?

m-andrew-albright:
If this does in fact solve the issue, it would be a huge win - we architected our application expecting `deactivate()` to be _guaranteed_ to be called - the workaround (using `unbind()`) does not work for us due to it not being guaranteed to be called before another module's `activate()`

EisenbergEffect:
@davismj Should I merge this PR? or does yours override this?

davismj:
@eisenbergeffect this can be closed

EisenbergEffect:
@jagonalez Thanks for starting things going on this. Looks like we've been able to expand on this fix and it will be merged soon.

Issue #486, opened 2017-04-03
baseUrl is incorrectly generated in child route when id parameter contains whitespace or other special characters


mzhang444:
**I'm submitting a bug report**

* **Library Version:**
1.2.1

**Please tell us about your environment:**
* **Operating System:**
OSX 10.11.13

* **Node Version:**
6.9.1

* **NPM Version:**
3.10.8

* **Browser:**
Chrome 56.0.29

* **Language:**
ESNext

**Current behavior:**
When "id" parameter is nested child route contains whitespace or other special characters that encodeURI will encode, the baseUrl in the child route are not calculated correctly.

**Expected/desired behavior:**

I have an application with following nested child routes defined:
users/:id
   - access-ctrl
       - accounts
       - accounts/:id
       - others
When account details is displayed, I have links to go back to user details, access control, accounts list page.

When account id is "abc", all the links are created correctly:
Account details page: http://localhost:9000/#/users/user1/access-ctrl/accounts/abc
User details page: http://localhost:9000/#/users/user1
Access control page: http://localhost:9000/#/users/user1/access-ctrl
Account list page: http://localhost:9000/#/users/user1/access-ctrl/accounts

When account is is "test 1", baseUrl becomes empty string for User details and account list page:
Account details page: http://localhost:9000/#/users/user1/access-ctrl/accounts/test 1
User details page: http://localhost:9000/#
Access control page: http://localhost:9000/#/access-ctrl
Account list page: http://localhost:9000/#/accounts

If I encode the whitespace to %20, then User details and and first level child route Access Control links are generated correctly, but not the nested Account list child route:
Account details page: http://localhost:9000/#/users/user1/access-ctrl/accounts/test%201
User details page: http://localhost:9000/#/users/user1
Access control page: http://localhost:9000/#/users/user1/access-ctrl
Account list page: http://localhost:9000/#/users/user1/accounts

* **What is the expected behavior?**
Whitespace and other special characters in "id" parameter should be handled properly. Users should not need to encode special characters. If special characters need to be encoded by the user, then the encoded value should be passed into the nested route instead of decoded value. In my case, even if I did encode the account id, it only works at level 1 child route. By the time it reaches level 2 child route, the id value got decoded and no longer works.

Looking at the source code, the issue is caused by this method. In the method, path is encode, while fragment is not encoded, therefore, this method can return empty string when path has special characters.

```
NavigationInstruction.prototype.getBaseUrl = function getBaseUrl() {
      var _this = this;
[...]

Issue #435, opened 2016-11-23
%(encoded value) in route parameter, then child routes does not calculate correctly


neolivz:
When you have an escaped value in route parameter (i.e. with %) and with child routes then the href generated values does not have a proper hierarchy.

Eg: [Modified Skelton Typescript Webpack](https://github.com/neolivz/skeleton-navigation/tree/master/skeleton-typescript-webpack).

[With %](http://localhost:9000/#/child-router/child-router/welcome/child-router/%23A/welcome
) All the parent routes are getting routed to top level ignoring the whole hierarchy.

<img width="848" alt="screen shot 2016-11-23 at 2 18 34 pm" src="https://cloud.githubusercontent.com/assets/754818/20555807/e704bac2-b189-11e6-9569-d13b05c21456.png">

[Without %](http://localhost:9000/#/child-router/child-router/welcome/child-router/#A/welcome
) When you don't have % in URL parameter it works fine
<img width="785" alt="screen shot 2016-11-23 at 2 19 04 pm" src="https://cloud.githubusercontent.com/assets/754818/20555838/106df09a-b18a-11e6-9023-edc2cfdcd595.png">

Check the URL generated.

EisenbergEffect:
@neolivz Would you be interested in submitting a PR to fix this issue? It sounds like it would be a nice contribution.

neolivz:
Sure :) I will look into it.

gheoan:
@davismj this can be closed.

PR #497, opened 2017-05-04
fix(router): getBaseUrl() to return correct URL for encodable chars

jagonalez:
This addresses issues: #435 and #486 

`fragment` may or may not be encoded but `path` will always be decoded. In order to compare `fragment` and `path` - `fragment` has to be decoded first.

This will cover any use-cases where fragment is not encoded, partially encoded or fully encoded.

This will affect `router.baseUrl` which will now be encoded. It will also affect the `href` parameter in the `router.navigation` array.

This means it may not be exactly the same as navigationInstruction.fragment for any of the routes if the URL isn't fully encoded to begin with.

### DIFF
--- a/src/navigation-instruction.js
+++ b/src/navigation-instruction.js
@@ -162,7 +162,7 @@ export class NavigationInstruction {
   * Gets the instruction's base URL, accounting for wildcard route parameters.
   */
   getBaseUrl(): string {
-    let fragment = this.fragment;
+    let fragment = decodeURI(this.fragment);
 
     if (fragment === '') {
       let nonEmptyRoute = this.router.routes.find(route => {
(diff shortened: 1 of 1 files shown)


davismj:
@EisenbergEffect has @jagonalez signed the CLA?

davismj:
I have extended this pull request to (a) include tests that fail before the fix is applied and (b) to include an additional fix that applies to the case where there is no `path`. I'm not sure the best way to to do this so I just opened a new pull request. @EisenbergEffect I recommend closing this in favor of #501 once confirmed that @jagonalez has signed the CLA.

EisenbergEffect:
@davismj I've confirmed the CLI signature.

PR #501, opened 2017-07-20
fix(navigation-instruction): getBaseUrl() to return correct URL for encodable chars

davismj:
This addresses issues #435 and #486 and extends pull request https://github.com/aurelia/router/pull/497.

fragment may or may not be encoded but path will always be decoded. In order to compare fragment and path - fragment has to be decoded first.

This will cover any use-cases where fragment is not encoded, partially encoded or fully encoded.

This will affect router.baseUrl which will now be encoded. It will also affect the href parameter in the router.navigation array.

This means it may not be exactly the same as navigationInstruction.fragment for any of the routes if the URL isn't fully encoded to begin with.

### DIFF
--- a/src/navigation-instruction.js
+++ b/src/navigation-instruction.js
@@ -182,7 +182,7 @@ export class NavigationInstruction {
     let path = this.params[wildcardName] || '';
 
     if (!path) {
-      return fragment;
+      return encodeURI(fragment);
     }
 
     return encodeURI(fragment.substr(0, fragment.lastIndexOf(path)));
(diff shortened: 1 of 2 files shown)


davismj:
@EisenbergEffect then this is good to be pulled in.

zewa666:
Tests are fine now, good to go i'd say

EisenbergEffect:

Vheissu:
I pulled this down locally in Windows 10 and tests all pass and the code appears to be working fine. Nice work @davismj this will be a nice one to get merged in.

PR #502, opened 2017-07-21
test(activation.spec): Remove obsolete activation test

davismj:
I removed a test that would erroneously fail. The test was checking that a plan with a childNavigationInstruction property that returned true or false on canDeactivate would continue or cancel respectively. The mock childNavigationInstruction was missing the required childRouter property, and so it was giving erroneous results.

The following test, 'with router and currentInstruction', actually includes coverage for the very same feature. Therefore, removing the erroneous test was sufficient.

### DIFF
--- a/test/activation.spec.js
+++ b/test/activation.spec.js
@@ -94,28 +94,6 @@ describe('activation', () => {
       expect(state.rejection).toBeTruthy();
     });
 
-    describe('with a childNavigationInstruction', () => {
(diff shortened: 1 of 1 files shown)


JeroenVinke:
Looks good!

EisenbergEffect:

Vheissu:
No complaints here, let's get this merged in @davismj

PR #503, opened 2017-07-21
fix(activation): invoke deactivate method on child modules

### LINKED ISSUE #132: Deactivate method isn't called when user is moving between two child routers.

davismj:
This pull request is based on and replaces https://github.com/aurelia/router/pull/496 by @jagonalez. 496 is working around an erroneous test. I have removed the test in https://github.com/aurelia/router/pull/502 and reimplemented this pull request without working around the test. 

Depends on #502 
Closes #132 #411 #496

### DIFF
--- a/src/activation.js
+++ b/src/activation.js
@@ -69,9 +69,7 @@ function findDeactivatable(plan, callbackName, list: Array<Object> = []): Array<
       }
     }
 
-    if (viewPortPlan.childNavigationInstruction) {
-      findDeactivatable(viewPortPlan.childNavigationInstruction.plan, callbackName, list);
-    } else if (prevComponent) {
+    if (prevComponent) {
       addPreviousDeactivatable(prevComponent, callbackName, list);
     }
   }
--- a/src/pipeline-provider.js
+++ b/src/pipeline-provider.js
@@ -35,8 +35,8 @@ export class PipelineProvider {
     this.container = container;
     this.steps = [
       BuildNavigationPlanStep,
-      CanDeactivatePreviousStep, //optional
       LoadRouteStep,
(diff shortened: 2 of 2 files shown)


davismj:
After a bit more testing, I've found that this doesn't work properly in the recursive router case.

davismj:
Seems like this is actually another existing bug and it wasn't introduced in this pull request.

davismj:
After discussing with the team, this is not a good PR.

PR #505, opened 2017-07-25
feat(activation): call can/Deactivate with navigationInstruction

### LINKED ISSUE #493: feature request: pass navigationInstruction to canDeactivate
**I'm submitting a feature request**

* **Library Version:**
1.3.0

* **Operating System:**
Windows 10

* **Browser:**
all

* **Language:**
TypeScript 2.3

**Current behavior:**
No arguments are passed to canDeactivate, so the requested route cannot be used to determine actions needed to be taken.

* **What is the expected behavior?**
Pass the navigationInstruction to canDeactivate so that the requested route can be inspected before taking action.

I want to confirm navigation only if the user requested a route outside of the current parent route. If the user navigates to a sibling route, I want to allow navigation without confirming. The scenario is a wizard, where the user can navigate between pages of the wizard before saving, but shouldn't be able to leave the wizard altogether.
This can be achieved by handling canDeactivate on the parent, but NOT if the activation strategy is "replace" or "invoke-lifecycle", because then canDeactivate is called on the parent when navigating between children too.

davismj:
This enables the consumer to inspect the navigation instructions and make a decision on whether or not a route can be deactivated based on the incoming navigation instructions.

Closes #493

### DIFF
--- a/src/activation.js
+++ b/src/activation.js
@@ -3,7 +3,7 @@ import {isNavigationCommand} from './navigation-commands';
 
 export class CanDeactivatePreviousStep {
   run(navigationInstruction: NavigationInstruction, next: Function) {
-    return processDeactivatable(navigationInstruction.plan, 'canDeactivate', next);
+    return processDeactivatable(navigationInstruction, 'canDeactivate', next);
   }
 }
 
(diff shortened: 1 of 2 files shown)


zewa666:
Is this a breaking change? I guess code out there that relies on the plan would have issues with the switch

davismj:
`processDeactivatable()` is internal API, so this is not a breaking change to the public API.

zewa666:
👍

EisenbergEffect:

davismj:
I've manually merged this pull request.

Issue #483, opened 2017-03-30
Unable to redirect with params


nym:
**I'm submitting a feature request**
Redirecting with parameters does not work in the same way that directing with query parameters do.

**Given:**
{route: "path/to/:id/old", redirect: "path/to/:id/new"} 

**Expected:**
/path/to/9000/old ->
/path/to/9000/new

**Actual:**
/path/to/9000/old ->
/path/to/:id/new

From the source, it appears that query parameters are passed along as part of redirectLocation, but parameters are not being replaced despite being available in instruction.params.

https://github.com/aurelia/router/blob/90b1183648b804d52a57090e3a9ba34982a22097/src/navigation-plan.js#L28

EisenbergEffect:
I agree, this would be a great feature. Any interest in working on a contribution to add it?

nym:
@EisenbergEffect sure- is there an approach you would recommend? My initial thought was that parsing / replacing params in the section I listed would make sense, but a) want to make sure that is correct and b) is there something that is designed to do this already that I should reuse instead of writing regex to replace by :param/ with ${param}/ ?

Finally, what are the requirements for getting a PR accepted? I am going to try to do this over the weekend and want to do it right the first time.

cc @devanp92

EisenbergEffect:
There are some existing APIs to help you out. You shouldn't need to mess with RegEx. Have a look at what happens with the basic route and you should be able to emulate the same thing.

balazsmeszegeto:
@nym I'm interested in creating a PR for this, if you'd need help

Alexander-Taran:
Can be closed

PR #515, opened 2017-08-29
feat(router): redirect with token parameters

jagonalez:
allow redirect to contain token parameters for router config

this closes feature request #483

### DIFF
--- a/src/navigation-plan.js
+++ b/src/navigation-plan.js
@@ -26,12 +26,18 @@ export function _buildNavigationPlan(instruction: NavigationInstruction, forceLi
   let plan = {};
 
   if ('redirect' in config) {
-    let redirectLocation = _resolveUrl(config.redirect, getInstructionBaseUrl(instruction));
-    if (instruction.queryString) {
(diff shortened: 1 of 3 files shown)


djensen47:
Any updates on this? Would be really nice to have.

EisenbergEffect:
@davismj Can you review this? It looks like an interesting feature.

davismj:
@jagonalez Sorry for the wait. Sincerely appreciate your hard work as always.

@djensen47 Thanks for your patience. Merged!

Alexander-Taran:
@jagonalez would you mind creating another PR for docs?
Or explain a bit on how it works.. so we could update the docs with a use case?

huochunpeng:
@jagonalez @davismj @EisenbergEffect 

This new design of redirect disabled the old usage of redirect as in #607
It also causes regression in #605

This looks like a very isolated change, can we revert this commit for now?

Need to revise the design and add test coverage to ensure not breaking #605 and #607.

Issue #472, opened 2017-02-22
Router CurrentInstruction is null in 1.2.0


allencoded:
Updated aurelia router to the latest 1.2.0 version and it seems to have a bug that isn't in the earlier versions.

` this.router.currentInstruction.getAllInstructions();`

Error
`Cannot read property 'getAllInstructions' of null`

Upon inspection router.currentInstruction is now null and I can't perform getAllInstructions() any longer due to that.

EisenbergEffect:
When/Where are you calling that api from?

allencoded:
I have a Breadcrumb class. Which currently works under aurelia router 1.1.1. Once navigation is successful it then calls the api endpoint:
this.router.currentInstruction.getAllInstructions(). 

In my attached:
```
this.subscription = this.ea.subscribe(
      'router:navigation:success',
      ::this.navigationSuccess);
```
```
navigationSuccess() {
    this.routes = this.router.currentInstruction.getAllInstructions();
}
```

At this point it is null only when I am using the latest version of the aurelia-router 1.2.0. Like I said though in my current 1.1.1 version it works well. 

I had copied Jeremy Danyow's response from here a long time ago when I first implemented this and our code is basically a direct mimic:
http://stackoverflow.com/questions/33689554/identify-the-current-route-in-aurelia

RichiCoder1:
@allencoded A temporary work around for this, if you haven't already found one, would be:
```
navigationSuccess(eventArgs) {
    const { instruction } = eventArgs;
    this.routes = instruction.getAllInstructions();
}
```
*should* work

EisenbergEffect:
I think we must have made a bug fix that caused a regression here. Is anyone interested in submitting a PR to fix the issue. I can make a quick turnaround on the release.

jagonalez:
@allencoded I'm not able to reproduce this error. Are you able to share a gist with code that creates the error?

mczp:
I think I just hit the same issue, code like this is triggering it:
```
@inject(MCNetRPC, Router)
export class FabricDetails {
  constructor(rpc, router) {
    this.rpc = rpc;
    this.router = router;
    this.loading = true;
    setTimeout(() => { this.load_fabric(); }, 0);
  }

  async load_fabric() {
    console.log(this.router);
    console.log(this.router.currentInstruction);
    //fid = this.router.currentInstruction.params.id;
    //console.log(`Loading fabric details FID: ${fid}`);
  }
}
```
In the console I'm getting:
[...]

PR #517, opened 2017-09-18
doc(router): fix data type of currentInstruction

gheoan:


gheoan:
I have not realised this is not intended: https://github.com/aurelia/router/issues/472

Issue #433, opened 2016-11-16
If the router has multiple routes using the same route and different hrefs then rowActive does not work.


glennpierce:
**I'm submitting a bug report**

* **Library Version:**
1.0.2

**Please tell us about your environment:**
* **Operating System:**
Linux Fedora23

* **Node Version:**
6.2.2

* **NPM Version:**
3.9.5

* **JSPM OR Webpack AND Version**
JSPM 0.16.32 | webpack 2.1.0-beta.17

2.1.0-beta.22

* **Browser:**
all | Chrome XX | Firefox XX | Edge XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView
Chrome 52

* **Language:**
all | TypeScript X.X | ESNext
typescript 2.0.0

**Current behavior:**

I have routes defined like

{ route: "shimmy_section/:id", href:'shimmy_section/yourcommunity', name: 'yourcommunity', moduleId: "./shimmy_section", nav:true, title: 'Your Community'},
      { route: "shimmy_section/:id", href:'shimmy_section/yourtravel', name: 'yourtravel', moduleId: "./shimmy_section", nav:true, title: 'Your Travel'},
      { route: "shimmy_section/:id", href:'shimmy_section/yourenergyandwater', name: 'yourenergyandwater', moduleId: "./shimmy_section", nav:true, title: 'Energy & Water'},
      { route: "shimmy_section/:id", href:'shimmy_section/yourneighbourhood', name: 'yourneighbourhood', moduleId: "./shimmy_section", nav:true, title: 'Your Neighbourhood'},

Ie they are handled by the same module hence I used href.
However, row.isActive does not work with this setup. The last item is always active

**Expected/desired behavior:**

The following example shows the issue

https://github.com/glennpierce/test

* **What is the expected behavior?**

rowActive should work with this setup.

* **What is the motivation / use case for changing the behavior?**
I wan't to show the active select in the nav bar.

a-jackson:
I'm see this issue as well. Any progress on it? I'd be willing to do the PR myself if someone could point me in the right direction

EisenbergEffect:
How should it determine which route is active?

a-jackson:
I would expect it to match the href.

I tried to look for an answer myself but got lost in the route-recognizer and its states. I can't tell where it matches a NavigationInstruction to a config. I know this lands somewhere in the route-recognizer.

[...]

PR #520, opened 2017-09-19
test(router): test multiple identical patterns

gheoan:
Fixes https://github.com/aurelia/router/issues/433.
Requires https://github.com/aurelia/route-recognizer/pull/30.

### DIFF
--- a/test/router.spec.js
+++ b/test/router.spec.js
@@ -239,6 +239,53 @@ describe('the router', () => {
         .then(done);
     });
 
+    describe('should match routes with same pattern based on href', () => {
+      it('', (done) => {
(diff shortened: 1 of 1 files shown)


gheoan:
This is not ready to be merged. The added test is failing and I am waiting feedback [here](https://github.com/aurelia/router/issues/433#issuecomment-330661265).

gheoan:
Is this a breaking change? I can't think of a scenario where someone had multiple routes with the same route pattern and expected all of the routes to actually navigate to the a single route (the first one returned by `route-recognizer`).

gheoan:
Decided to move the logic to route-recognizer and leave only some tests here.

PR #530, opened 2017-10-04
test(router): fix promise catch logging

gheoan:


EisenbergEffect:
@martingust This looks like fixes to tests. Can you do a review?

Issue #525, opened 2017-09-27
Navigating to non-empty child route fails when parent route is empty


gheoan:
**I'm submitting a bug report**
* **Library Version:**
1.3.0

**Please tell us about your environment:**
* **Browser:**
Tested in latest Chrome, Firefox

* **Language:**
all

**Current behavior:**
Consider this configuration:
```js
  configureRouter(config) {
    config.mapRoute({
      route: '',
      moduleId: 'child',
    });
  }
```
```js
// child.js
export default class {
  configureRouter(config) {
    config.mapRoute({
      route: 'test',
      moduleId: 'test',
    });
  }
}

```

Navigating to `/test` fails with error route not found.
Here is a demonstration: https://gist.run/?id=929bcf7c2c64fbf6c9e3f2b6f7e797ab.
Seems like this is similar to https://github.com/aurelia/router/issues/27.

**Expected/desired behavior:**

* **What is the expected behavior?**
`/test` should match the only child route from the example above.

gheoan:
Here is a reproduction using the latest aurelia libraries: https://github.com/gheoan/aurelia-router-issue-525.

Alexander-Taran:
http://aurelia.io/docs/routing/configuration#redirecting-routes

> The redirect is particularly useful when you have an "empty" route pattern (such as the first route above) that maps to a component with a child router. In this case, create a non-empty route and then redirect the empty route to the non-empty route (as above). This will enable the child router to consistently match child routes without getting confused in scenarios where the empty route was matched.

seems like all the threads about empty parent routes end there


maybe can be closed

Alexander-Taran:
@davismj  (-:

PR #532, opened 2017-10-05
fix(router): add empty route to child recognizer

gheoan:
Fix https://github.com/aurelia/router/issues/525.


EisenbergEffect:
@davismj Can you review this?

Issue #534, opened 2017-10-14
Feature query: stateful routes (module + viewport)


jwx:
I've got code for stateful routes, meaning that if you do this
```
{ route: 'dashboard', name: 'dashboard', viewPorts: { 'default': { moduleId: 'dashboard', stateful: true } } }
```
the same instance of the dashboard module will be presented in the default viewport everytime a stateful route specifies it after the initial time. The view/viewmodel will remain loaded even when navigating away to another route, meaning that state such as list sorting, entered values, positions on google maps etc will be the same when coming back to it. It also means that the viewmodel will be able to listen to and act upon events from the event manager, signals and so on.

So to clarify, this is not caching for views/viewmodels, it's stateful views/viewmodels that are never unloaded.

If this is of interest to anyone else and has any chance of being accepted into the router I could put together a PR.

jwx:
Is this something you want in? @EisenbergEffect @bryanrsmith

TonyLugg:
I like it!

davismj:
@jwx Isn't this solved by making your view model a app singleton?

davismj:
@EisenbergEffect see here

jwx:
@davismj @EisenbergEffect Making the view model a singleton only preserves the view model state. State details not bound to the view model is lost, for example an unbound scroll position in a list or unbound position and zooming in a google map. Basically, this is completely cached views and so on.

davismj:
Good info. I'll review again.

I'd like to pull this in out of the context of the optional and default view ports. This seems like a totally separate item.

bigopon:
I'd like to add in that making something `singleton` is not always viable. Sometime the element used at the leaf - route may end up being a real element in another context, thus making them singleton will not allow this scenario. My use case is I have a complex editor as a route, but then in another route, I reuse the element in a dialog.

Alexander-Taran:
Don't like the seeming hackiness of it. Kinda shoehorning.
you actually can achieve it with optional viewports now i think and a bit of css magic.

jwx:
@Alexander-Taran See https://github.com/aurelia/router/pull/538.

jfstephe:
@bigopon @jwx - any news on this or #538 ?

Issue #528, opened 2017-10-03
Cancel navigateBack from canDeactivate


TonyLugg:
**I'm submitting a bug report**

* **Library Version:**
aurelia-router: 1.3.0

* **Operating System:**
Windows 10

* **Node Version:**
6.11.3

* **NPM Version:**
5.3.0

* **JSPM OR Webpack AND Version**
Webpack 3.5.5 (using CLI)

* **Browser:**
all

* **Language:**
TypeScript 2.5.2

**Current behavior:**
I load the app, which loads "home" route/view.  Press a button and use router.navigate("second") to go to second page.  Enter some data and press back button (on page) which executes router.navigateBack().  Because data has changed, display a message box (Aurelia dialog) from canDeactivate() to ask for user confirmation (same behavior with window.confirm()), and return false from canDeactivate().  Navigation stops.  Repeat from pressing back button (on page), backward navigation continues and navigates right out of the app as if it has navigated back twice.

**Expected/desired behavior:**
Each time false is returned from canDeactivate, navigateBack() should be stopped.

If I press a different button and perform navigate("someOtherRoute") and return false from canDeactivate it works correctly with multiple navigation attempts.

TonyLugg:
Just wondering if this has any attention.  It is causing me a bit of grief.

davismj:
@TonyLugg Thanks for your detailed report.

Running gist here: https://gist.run/?id=f4a7fa39c9d82f5de0af3b0bc8e01850

gama410:
Hello!

Any update on this bug?

davismj:
I'm going to want to start looking at this once for 1.6.0, potentially a March release if all goes well.

gama410:
I'm not familiar with the implementation details but wouldn't it be logical to call history.back only after candeactivate returns true? It seems weird to me that we navigate back before checking if we are allowed to.

davismj:
I've found a number of related bugs. Specifically, if you start at `#/first` and navigate to `#/second` via `<a href="#/second">`, and if `first` returns `false` from `canDeactivate()`, a new history item would be appended for the current route.

Some early tests show this solving this issue and most of the related issues I've uncovered:

```

  function restorePreviousLocation(router) {
    var previousLocation = router.history.previousLocation;
    if (previousLocation) {
      Promise.resolve().then(() => history.go(router.isNavigatingBack ? 1 : -1));
    } else if (router.fallbackRoute) {
[...]

PR #538, opened 2017-10-17
feat(router): stateful routes

jwx:
Adds stateful routes so that route configurations can specify a module or a module in a viewport as `stateful: true`. A stateful module that's loaded in a viewport is never unloaded when navigating away, it's just not shown, and is displayed with the same state whenever a route places it in the same viewport again.

Depending on aurelia/templating-router#64 and aurelia/router#536.
Closes aurelia/router#534.


EisenbergEffect:
@davismj This is a pretty highly requested feature. The PR has quite a few changes, so needs careful review.

davismj:
@EisenbergEffect I thought I had replied to this one. What does this solve that a `@singleton` view model doesn't?

oneillci:
@davismj @jwx  Can I ask what's the status of this pull request? This looks like a feature that would be very useful for my application

davismj:
@oneillci I need a bit of time to review this. I'll keep this on my radar for an April release.

jfstephe:
@jwx - Can I just check something? If I had a route:

person/:id/details

and I navigated through 100 people, it would have 100 hidden people waiting in the DOM(view)/in-memory(view model). Is this correct?

By 'hidden' do you mean 'display:none' in css or removed from the DOM?

Apart from increase memory demands are there any drawbacks of this feature?

Do you still get the same 'activate' etc events in the 'detail' VM?

Is there a way to clear the cache, if you so wanted?

TonyLugg:
Was this every implemented?  I am building an app that has a geographic map and I want the map to be retained in memory when I navigate away and return.

TonyLugg:
And this one too please :)
https://github.com/aurelia/router/issues/528

oising:
Bump.

jfstephe:
@davismj @EisenbergEffect - any news on this? Any eta?

jfstephe:
@jwx @davismj @EisenbergEffect - any news on this? Any eta?

fkleuver:
Apologies for keeping everyone waiting on this for so long.

At the moment our priority is with getting Aurelia 2 to alpha. That is not to say that we're abandoning v1, but I strongly believe it to be in the majority of the community's best interest that we *first* get v2 to a state where we can comfortably recommend it for production, before adding any new features to v1.

This feature as well as almost all other requested features are in fact already implemented and well-tested in v2. The bulk of the remaining work in v2 is further improving test coverage, and maximizing backwards compatibility with v1 to ease migration.

The original plan was to get v2 done in Q1 this year. Due to various circumstances we won't make that, so we're trying to make sure it will at least be Q2 at the latest, and to maximize the odds of accomplishing that we need to focus all effort on v2 for the time being. After that, the idea is to circle back to v1 and backport some tests and features where possible, bringing v1 and v2 closer to each other to make migration even easier.

jfstephe:
@fkleuver - many thanks for the detailed reply, Looking forward to v2! :-)

PR #539, opened 2017-10-18
feat(router): add params to navigate()

jwx:
Adds a `params` parameter to the `navigate()` function, so that
```
router.navigate('#/main/:id/details', undefined, { id: 123, name: "onetwothree" })
```
results in navigation to
```
'#/main/123/details?name=onetwothree'
```

In other words, found named params are placed in the url, not found ones are added to the querystring.

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -1,6 +1,7 @@
 import {RouteRecognizer} from 'aurelia-route-recognizer';
 import {Container} from 'aurelia-dependency-injection';
 import {History} from 'aurelia-history';
+import {buildQueryString, parseQueryString} from 'aurelia-path';
 import {NavigationInstruction} from './navigation-instruction';
 import {NavModel} from './nav-model';
 import {RouterConfiguration} from './router-configuration';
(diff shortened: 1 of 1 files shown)


jwx:
Is this within the scope of the router? (At least it's a quite small addition.) @EisenbergEffect @bryanrsmith @davismj

davismj:
Is there anything this solves that doesn't currently work with `router.navigateToRoute`?

```javascript
router.navigateToRoute('main-details', { id: 123, name: 'onetwothree' });
```

jwx:
@davismj `router.navigate` doesn't require route configuration.

davismj:
@jwx Why would you want to navigate to a route that isn't configured?

davismj:
I'm not going to close this. I see a great deal of value in having a utility function that generates a URL with query string parameters as a first-class citizen in Aurelia. I'm just not sure it goes here. It might!

bigopon:
I think this feature is desirable. Currently I don't see how `navigateToRoute` of child Router can help navigate to routes in different route branches, so one may end up rolling their own, like the feature this PR adds.

jwx:
@davismj Any chance of getting this merged?

CasiOo:
@bigopon is right. This PR will be very useful when dealing with navigation with child routes

Issue #479, opened 2017-03-08
Feature request: passing state when navigating


jwx:
**I'm submitting a feature request**

A somewhat common scenario (at least for me) is the desire to pass data/a state when navigating from one view to another. This state could for example be the name of previous view, selected sort order or some json data that's been fetched. The solution (for me, most of the time) is to pass the state through an injected singleton and while this state singleton is useful for global state or when additional functionality is required, it just feels cumbersome for passing state when navigating.

I propose adding a new parameter, _state: any_, to _navigate_ and _navigateToRoute_ that will be transfered into _currentInstruction_ so that the state is available in the new view's _activate_ function.

In addition, I propose expanding _navigate_ to be able to take an url in the same format that is used when configuring routes so that it's possible to get url generation in _navigate_ with the aid of an additional new parameter, _parameters: any_ (mirroring the parameters for _navigateToRoute_).

I'm currently doing variations/patches of the above in my projects so if this is something you want, I'd be willing to create the PR.

jwx:
@EisenbergEffect

EisenbergEffect:
I've generally been apposed to this because it means that there's some aspect of the history which will break when the browser is refreshed or when a link is shared. Can you talk more specifics about the scenario where you don't think these issues would be a concern?

jwx:
It's intended for data that's either non-essential for functionality or that's fetched if not present. I've seen the question on how to pass <data or object> into the next view on gitter a couple of times and the suggestions tend to be _state singleton_ (which presents the same problems when it comes to links and refresh).

An example I have is from a mobile app where selecting an item in a list of items navigates to /item/:id/ and passes the item data object from the list. The target view then initiates a fetch based on the id, but presents the item based on the passed object, if there is one, while waiting for the reply. (In some cases, it didn't even perform the fetch if the data was passed along.) It made a difference for the user experience.

(And then, of course, there's the, albeit very few, times when one doesn't care about supporting external links or page refresh.)

shanonvl:
Hi @jwx -- I've hit this myself when developing apps and wanted to throw this out for consideration in lieu of adding this functionality to Aurelia.

While it's still a simple case, IMO this is the exact use case for a logical _data layer_ ie. separation of concerns wrt data and presentation.

If I’m following correctly, in your activate:

```js
activate(routeConfig, params) {
  const id = params.id,
     allItems = params.allItems,  // assume this magically gets wired to [ { id: 1 }, { id:2}, { … ], { id:10 } ]
[...]

PR #540, opened 2017-10-18
feat(router): add state parameter to navigation

jwx:
Adds a `state` parameter to the `navigate` and `navigateToRoute` functions. The parameter is transfered to a `state` property on the NavigationInstruction.

Depends on aurelia/history-browser#30.
Closes aurelia/router#479.


davismj:
As #479 is closed, I'm going to close this too.

Issue #522, opened 2017-09-23
Router Doc for Prerender Step has an Error


TonyLugg:
In the Router documentation http://aurelia.io/hub.html#/doc/article/aurelia/router/latest/router-configuration/8, A Prerender Step, the example is missing '=>'.  The doc has:

`
    let step = {
      run: (navigationInstruction: NavigationInstruction, next: Function): Promise<any> {
        return next();
      }
    };
`

...and it should be:

`        let loStep = {
            run: (navigationInstruction: NavigationInstruction, next: Function): Promise<any> => {
                return next();
            }
        };`

PR #542, opened 2017-10-22
doc(router-configuration): refactor code example

gheoan:
https://github.com/aurelia/router/issues/522

### DIFF
--- a/doc/article/en-US/router-configuration.md
+++ b/doc/article/en-US/router-configuration.md
@@ -29,7 +29,7 @@ To use Aurelia's router, your component view must have a `<router-view></router-
 </code-listing>
 
 <code-listing heading="Basic Route Configuration">
-  <source-code lang="ES 2015/2016">    
+  <source-code lang="ES 2015/2016">
     export class App {
       configureRouter(config, router) {
         this.router = router;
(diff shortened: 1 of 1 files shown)


doc/article/en-US/router-configuration.md
```diff
@@ -403,12 +403,12 @@ A pipeline step must be an object that contains a `run(navigationInstruction, ne
...
   </source-code>
   <source-code lang="TypeScript">
-    import {Redirect, NavigationInstruction, RouterConfiguration, Next} from 'aurelia-router';
+    import {RouterConfiguration, NavigationInstruction, Next, Redirect} from 'aurelia-router';
 
     export class App {
       configureRouter(config: RouterConfiguration): void {
-        config.title = 'Aurelia';
-        config.addAuthorizeStep(AuthorizeStep);
+        const step = new AuthorizeStep;
```
davismj: What happened here?

gheoan: First, the title configuration is removed because it is not needed for this example.
Second, the `addAuthorizeStep` function expects as argument an object with a `run` method. The `AuthorizeStep` is a class so it does not satisfy that requirement bacause it contains a step method under the prototype (`AuthorizeStep.prototype.run`). Since, an instance of the class is created which is what `addAuthorizeStep` expects.

davismj: I see. This does work, but I think in the docs we should prefer relying on the Aurelia DI. That is, we should pass the AuthorizeStep class as a key and let Aurelia DI instantiate it. In an application it isn't significant, but everyone is going to copy paste this code so its best we stick to best practices and conventions.

davismj:
@gheoan thanks for this!

PR #544, opened 2017-10-23
fix(navigation): add NavigationResult type

AshleyGrant:
type NavigationResult = PipelineResult | boolean;
ref: https://discourse.aurelia.io/t/checking-for-candeactivate-before-navigatetoroute/51/2

### DIFF
--- a/src/interfaces.js
+++ b/src/interfaces.js
@@ -188,3 +188,18 @@ interface ActivationStrategy {
   */
   replace: 'replace';
 }
+
+/**
+* The result of a pipeline run.
+*/
+interface PipelineResult {
+  status: string;
+  instruction: NavigationInstruction;
+  output: any;
+  completed: boolean;
+}
+
+/**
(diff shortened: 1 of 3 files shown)


davismj:
@AshleyGrant per discourse, should this then be `Promise<PipelineResult>`?

AshleyGrant:
I want able determine that, so I made it both.

src/interfaces.js
```diff
@@ -188,3 +188,18 @@ interface ActivationStrategy {
...
+  status: string;
+  instruction: NavigationInstruction;
+  output: any;
+  completed: boolean;
+}
+
+/**
+* The result of a router navigation.
+*/
+type NavigationResult = PipelineResult | boolean;
```
davismj: Shouldn't this be 

`type NavigationResult = Promise<PipelineResult | boolean`?

davismj:
Thanks @AshleyGrant

PR #551, opened 2017-11-12
feat(router): Support optional and default view ports

davismj:
Before, all viewports had to be specified in every route. This feature enables omitting a viewport or specifying a default viewport.

Any viewport configuration with `moduleId: null` will use the default viewport configuration. Initially, the default viewport configuration is an empty view and view model pair. You can override this by passing a viewport configuration to the `useDefaultViewPort()` method on `RouterConfiguration`.

```javascript
configureRouter(config, router) {
  // this means that whenever the menu viewport is populated with { moduleId: null } in the router, it will populate with the 'defaultMenu' module.
  config.useDefaultViewPort({ menu: { moduleId: PLATFORM.module('defaultMenu') } });
}
```

If a route is missing a viewport configuration, then that route will remain unchanged. If there is no history, then that route will use the default viewport configuration. This allows you to specify routes that operate only on one viewport at a time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "karma-sourcemap-loader": "0.3.7",
     "merge2": "^1.0.2",
     "object.assign": "^4.0.4",
-    "require-dir": "^0.3.0",
+    "require-dir": "^0.3.2",
     "run-sequence": "^1.2.2",
     "through2": "^2.0.1",
     "typedoc": "^0.4.4",
--- a/src/navigation-plan.js
+++ b/src/navigation-plan.js
@@ -1,5 +1,5 @@
-import {Redirect} from './navigation-commands';
-import {_resolveUrl} from './util';
+import { Redirect } from './navigation-commands';
+import { _resolveUrl } from './util';
 
 /**
 * The strategy to use when activating modules during navigation.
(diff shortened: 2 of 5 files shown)


bigopon:
It looks like we are close to having a registry of view port - view model. Is there any chance this will enable the scenario where we can specify view port by a name ?

Like `config.registerViewport({})`

davismj:
@bigopon not sure what you mean. This enables creating routes like this:

```javascript
configureRouter(config, router) {
  config.map({

    // only modify one viewport
    { route: 'left/a', viewPorts: { left: { moduleId: 'a' } } },

    // modify two viewports
    { route: 'both/a', viewPorts: { left: { moduleId: 'a' }, right: { moduleId: 'a' } } },

    // modify two viewports, leave one empty
    { route: 'left/a/right/empty', viewPorts: { left: { moduleId: 'a' }, right: { moduleId: null } } }

  })
}
```

bigopon:
Wow yeah it's terribly unclear. I meant if we can some sort register a default view port, maybe create a registry, and then in the root `AppRouter`, we can do

```js
// config.registerViewPort('default', 'Module that will be on every page but no need to care about')
config.registerViewPort('default', 'nav-bar');
```

And then in the child router, we can refer that view port from child view port registry, or if child view port registry doesn't have it, look up on the parent register, similar to aurelia Container. ... I still don't feel I put it clear enough so let me know, I can give it another go @davismj  😆

davismj:
I think I understand, you want to programatically access a view port? That's not really related to this. Add a new issue and we'll discuss there.

Issue #527, opened 2017-10-03
Aurelia Router should support `ignoreUnknownRoutes`-like feature


PWKad:
The router for Durandal included an option to ignoreUnknownRoutes IIRC.  If a link is clicked on and Aurelia doesn't have a matching route the request should be able to continue on to the server by having Aurelia completely ignore unknown routes.  

Perhaps this could take a function to override the default behavior so the user can configure more thoroughly.

davismj:
@PWKad A couple questions:

- If navigation to /my/url occurs and /my/url isn't configured, the request falls through to a normal navigation request by the browser for /my/url, correct?
- This shouldn't have any effect unless on a pushState app, correct?
- What would the "default behavior" be? What could you override it with?

PWKad:
Right now the router hijacks any request that comes through when push state is active at the same base URL IIRC and the only way to allow the server to handle the request so the router doesn't swallow it is to map unknown routes and return a fake promise that never resolves, which is ugly.  

Instead the router should allow an option to not throw if a matching route isn't found which could be called something like `ignoreUnknownRoutes` I thought this was a feature from Durandal IIRC but I may be mistaken but it's needed for push state apps where the request might be serviced by the backend instead.

pndewit:
@PWKad Thanks, your workaround works! 👍 

If anyone is interested, I use the following piece of code inside my `configureRouter(routerConfiguration, router)` function:

```
routerConfiguration.mapUnknownRoutes(() => {
    if (location.pathname.startsWith('/my-route')) {
        return './app/not-found.html';
    } else {
        // REALLY ugly workaround (introduces memory leak) to ignore unknown routes outside "/my-route"
        this.dummy_promise = this.dummy_promise || new Promise(() => {});
        return this.dummy_promise;
    }
});
```

`this.dummy_promise` is used (as @PWKad explained) as fake promise which never resolves. This fake promise is then reused for next calls to prevent the app from creating a new Promise object for each navigation call to an unknown route. This will only create another pointer in memory to that same fake promise (which of course is still a memory leak), but only a minor one.

davismj:
@mgraf1

#558 doesn't seem to solve this issue

(1) created a push state enabled application
(2) added a link to a local path file (build/paths.js)
(3) click the link, the router quietly navigates to build/paths.js, but the request does not fall through to the server

PWKad:
[...]

Issue #428, opened 2016-10-25
Ability to ignore a specific URL/link


fracz:
**I'm submitting a feature request / opening a discussion**

I have a router configured with HTML5 push state enabled. When I try to click links that download static files such as:

``` html
<a href="/files/some-file.xlsx">Download</a>
```

the router complains:

> ERROR [app-router] Error: Route not found: /files/some-file.xlsx

I have solved this by adding `target="_top"` to the link but I'm not sure if this is a correct solution.

``` html
<a href="/files/some-file.xlsx" target="_top">Download</a>
```

I have found that [competition solves this by adding `target="_self"` to such links](http://stackoverflow.com/a/17407179/878514). It also [ignores all the links starting with slash](https://docs.angularjs.org/guide/$location#html-link-rewriting).

Maybe aurelia-router should implement similar behavior?

EisenbergEffect:
@bryanrsmith What do you think about this?

davismj:
@fracz What happens if you use `target="_blank`. That's probably what I would recommend anyway, so I'd be curious to know if it would work.

EisenbergEffect:
Closing since no followup from OP in over 3 weeks.

benstevens48:
@EisenbergEffect  I also think there should be some way of preventing the router from processing a link (i.e. `<a href="...">...</a>`). 

In addition to the OP's example, this is needed in order to link from part of a site that uses Aurelia client-side routing to another part of a site which does not use client-side routing (and possibly does not use Aurelia).

Ignoring links with the attribute `target="_self"`, or indeed any target attribute, seems to be a standard way of doing this (see Angular 1, page.js). An alternative would be to create a directive to do this, but I definitely think it should be built-in to Aurelia. Also, if not the case already, the router should also ignore links with the download attribute.

davismj:
@benstevens48 @fracz Take a look at the feature request here: https://github.com/aurelia/router/issues/527

I think this is probably the most robust way to solve these kinds of problems. I have this on my radar for a 1.6.0 release of the router.

benstevens48:
Oh, yes, that also works, and is probably good enough, although it's not exactly the same. For example, you might still want to have a custom 404 not found page in your Aurelia app (so not ignore unmatched routes), whilst still specifically creating some links that should result in requests to the server.

davismj:
Hm. Good, really good point, @benstevens48 

[...]

PR #558, opened 2017-12-10
feat(router config): ignoreUnknownRoutes option

mgraf1:
Implements the ignoreUnknownRoutes option mentioned in issue https://github.com/aurelia/router/issues/527.

### DIFF
--- a/src/app-router.js
+++ b/src/app-router.js
@@ -43,8 +43,10 @@ export class AppRouter extends Router {
     return this._createNavigationInstruction(url)
       .then(instruction => this._queueInstruction(instruction))
       .catch(error => {
-        logger.error(error);
-        restorePreviousLocation(this);
+        if (!this.ignoreUnknownRoutes) {
+          logger.error(error);
+          restorePreviousLocation(this);
+        }
       });
   }
 
--- a/src/router-configuration.js
+++ b/src/router-configuration.js
@@ -12,6 +12,7 @@ export class RouterConfiguration {
   title: string;
   unknownRouteConfig: any;
   viewPortDefaults: any;
+  ignoreUnknownRoutes: boolean;
 
   /**
   * Adds a step to be run during the [[Router]]'s navigation pipeline.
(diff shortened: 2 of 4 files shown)


davismj:
@EisenbergEffect I've done some simple testing and I've found that this doesn't seem to solve the issue it was trying to address. In any case, I think we shouldn't release this until its tested and verified, and I'm looking to get a release of the latest changes out in the next week. Shall I revert this using the GitHub revert feature, or manually in Git?

EisenbergEffect:
GitHub revert is fine.

PWKad:
@davismj https://github.com/aurelia/router/pull/561/files this shows how it should work although that's not a final version, as a note there are a handful of failing specs right now but they were failing before I opened that PR.

PWKad:
In a nutshell if you have pushState enabled you should still be able to create an href that is not swallowed by the router (they all currently are) and instead you should be able to trigger navigation that the router ignores.  I think the existing PR simply swallowed the error but we need to propagate the route up which the PR I linked above addresses but is probably a work in progress, let me know if that makes sense.  Also if we are not releasing until Tuesday please do not revert a commit without giving the author a chance to fix the issue.

davismj:
@EisenbergEffect opened, I'll pull it in

@PWKad its not that I don't want him to finish it, its that I'd rather not rush it, I'd rather have the time to test these things fully before they get pulled in.

That said, please have a look at https://github.com/aurelia/router/issues/428. I think if we take a step back and ask what kind of use cases we need to support we can improve upon the lessons of Durandal. 

My suggestion there was, instead of ignoreUnknownRoutes, which would essentially break mapping unknown routes to a custom 404 page, is having a route configuration that has some sort of pass through flag.

pndewit:
@davismj The changelog of v1.5.0 says it got pulled in. Just checked and it is not, can you remove the entry from the releases changelog?

davismj:
Thanks, done.

PR #562, opened 2017-12-16
Revert "feat(router config): ignoreUnknownRoutes option"

davismj:
Reverts aurelia/router#558


Alexander-Taran:
revert-558-ignore-unknown-routes because it did not work (c) @davismj

Issue #476, opened 2017-03-05
Router events fired out of order when using layouts


PWKad:
**I'm submitting a bug report**

* **Library Version:**
Latest

**Please tell us about your environment:**
* **Operating System:**
OSX 10.x

* **Node Version:**
6.2.0

* **NPM Version:**
3.8.9

* **JSPM OR Webpack AND Version**
au-cli latest

* **Browser:**
all | Chrome XX | Firefox XX | Edge XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView

* **Language:**
all

**Current behavior:**

When using layouts, router events are triggered out of order

**Expected/desired behavior:**

Router events should fire at the same time when using layouts as when not

Open the gist and look at console to see order of events.
https://gist.run/?id=250495c8d6233c6dd1412d2d954479e0

davismj:
I didn't even know layouts were a thing. What do layouts do that the "shell", in this case app, doesn't?

huochunpeng:
Good new @davismj #571 fixes this bug.
With the fix, the events now fired same with or without layout.

jordan-ware:
I think the issue here is that you have `activate` promises in both the layout and the module, but the router navigation instruction finishes before both of these promises resolve. 

They seem to prevent the layout attaching to the DOM, but they should also hold up the routing until the promises resolve like in a normal parent/child router scenario.

I don't have a gist yet but have observed it in my own project... Seeing a blank screen because routing is finished but the layout activate is still waiting to be resolved and isn't attached to the DOM.

PR #571, opened 2018-01-10
fix(navigation-instruction): fix the order of delayed swaps.

### LINKED ISSUE #548: Missing parentOverrideContext when a nested child route is loaded at first time

* **Library Version:**
1.4.0

* **Operating System:**
OSX 10.13

* **Node Version:**
8.9.0

* **NPM Version:**
5.5.1

* **JSPM OR Webpack AND Version**
NA

* **Browser:**
all

* **Language:**
ESNext

**Current behavior:**
When a nested child route is loaded at first time, the parentOverrideContext is missing.

https://gist.run/?id=24bddb1ca8e502bcac280fbb3368dd30

This demo setup a nested route, two routes at top level, plus two nested routes in one of the top level page.

The top app.js module defined a variable `face` with value ":-)". Due to all components loaded by aurelia-router have access to app.js component through overrideContext chain, the app is able to display the face in every leaf pages.

But there is a problem, when nested page "/two/one" is loaded first time, the face is missing, but if you visit "/two/two" then visit "/two/one" again, the face comes back.

A debug log in `two/one.js` prints bindingContext and overrideContext in `bind()` callback, it reveals the parentOverrideContext was missing when page "/two/one" was loaded first time, it comes back to normal when it was loaded second time.
screen shot 2017-11-06 at 12 50 50 pm

* **What is the expected behavior?**
The first load of nested child route should have correct parentOverrideContext set.

I was trying to have a fix by looking into router implementation about the bindingContext manipulation during route activation, but could not find where the related code is. I would appreciate someone from Aurelia team could give me some hint, I would like to attempt to fix it.

3cp:
Previously, when a child route is loaded at first time, the swap() of child router-view is called before the swap() of parent router-view. The result is when binding the inner view, it could not get correct binding scope since its owningView was not bound yet. This fix makes sure the swaps are executed in the right order.

Closes #548, #476

### DIFF
--- a/src/navigation-instruction.js
+++ b/src/navigation-instruction.js
@@ -12,7 +12,9 @@ interface NavigationInstructionInit {
 
 export class CommitChangesStep {
   run(navigationInstruction: NavigationInstruction, next: Function) {
-    return navigationInstruction._commitChanges(true).then(() => {
+    return navigationInstruction._commitChanges(true).then(delayJobs => {
+      return delayJobs.reduce((chain, job) => chain.then(job), Promise.resolve());
+    }).then(() => {
       navigationInstruction._updateTitle();
       return next();
     });
@@ -202,7 +204,7 @@ export class NavigationInstruction {
     router.refreshNavigation();
 
     let loads = [];
-    let delaySwaps = [];
+    let delayJobs = [];
 
     for (let viewPortName in this.viewPortInstructions) {
       let viewPortInstruction = this.viewPortInstructions[viewPortName];
@@ -217,7 +219,7 @@ export class NavigationInstruction {
           loads.push(viewPortInstruction.childNavigationInstruction._commitChanges(waitToSwap));
         } else {
           if (waitToSwap) {
-            delaySwaps.push({viewPort, viewPortInstruction});
+            delayJobs.push(() => viewPort.swap(viewPortInstruction));
           }
           loads.push(viewPort.process(viewPortInstruction, waitToSwap).then((x) => {
             if (viewPortInstruction.childNavigationInstruction) {
@@ -232,10 +234,16 @@ export class NavigationInstruction {
       }
     }
 
-    return Promise.all(loads).then(() => {
-      delaySwaps.forEach(x => x.viewPort.swap(x.viewPortInstruction));
-      return null;
-    }).then(() => prune(this));
+    const promise = loads.reduce((chain, load) => {
+      return chain.then(() => load).then(_delayJobs => {
+        if (_delayJobs && _delayJobs.length) delayJobs.push.apply(delayJobs, _delayJobs);
+      });
+    }, Promise.resolve());
+
+    return promise.then(() => {
+      delayJobs.push(() => prune(this));
+      return delayJobs;
+    });
   }
 
   _updateTitle(): void {


huochunpeng:
I am not sure how to write the test code to cover this issue, would appreciate some help.
I tested it against my production app which has pretty nasty deep routes.

davismj:
Although your approach to this problem is right, I'm almost positive this is going to break some stuff in the framework as well as in live applications as this will likely fire the lifecycle callbacks in reverse. I'd need to investigate to confirm.

huochunpeng:
I agree, fundamental change. This needs thorough test. BTW, I need some help to add test coverage against this.

So far, it does not break my apps. My apps have quite complex router config. This is a good sign.

huochunpeng:
From what I can remember, I think I traced lifecycle callbacks.
With this fix, lifecycle callbacks were fired in expected order just like without router (I mean comparing to nest inner component directly in outer component).

Outer bind
Inner bind
Inner attached
Outer attached

Without this fix, when outer and inner were loaded by router, the lifecycle callbacks were not fired in the above sequence.

huochunpeng:
If this does restore the lifecycle callback sequence to intended way, I expect instead of breaking something, this should fix some other open issues.

Alexander-Taran:
can this also fix this: https://github.com/aurelia/router/issues/410 ?

huochunpeng:
@Alexander-Taran well, you can help to fix 👍 . You always look like got more time than others 😄

huochunpeng:
I vaguely remember I tried that one. If I did, the result is negative.

Alexander-Taran:
Well they seemed kinda related.. I was just curious.. Does your PR ensures that parent router is completely initialized before child.. 
so that apis mentioned in the issue behave consitently on deeplink & navigation

And as you can see, I mostly spend my time reading and answering.. (-:
not coding.. somebody has to

huochunpeng:
There are few issues around Redirect and Promise, I produced them. It looks like some edge cases not related to this issue, I didn't spend more time as I don't use Redirect that often and I am lazy.

fkleuver:
If there is a risk of breaking stuff, would it instead be an idea to make this behavior available via an option like `config.options.swapOrder = swapOrder.default | swapOrder.topDown`?

huochunpeng:
Negative, if this one breaks stuff, it means it has bug. Doesn't mean existing one has no bug. We probably should not ask user do you want bug1 or bug2?

fkleuver:
LOL.

I think you're missing the point though. If there was a bug in the implementation from the very start, especially a subtle one like this which can creep into the architecture, then consumers might have built the application in a way that works around it.

Then, if you fix that bug - and your implementation has **no bugs**, then that in itself might break applications as described above.

So some people might actually want to opt-in to the **correct** behavior or have a chance to test/migrate to it without it being the only option. That's just my opinion though.

huochunpeng:
Thx, make sense. Until my fix break something.

Alexander-Taran:
@huochunpeng how about this one?
https://github.com/aurelia/framework/issues/810

PWKad:
Leaving a bug in because it was written early means that everyone who chooses to use the feature is going to run in to an inconsistency that they have to spend hours looking for work arounds for.  If this fixes the issue with layouts and order of what happens it should be merged and released as a "breaking" change.

fkleuver:
Still, wouldn't making this an **option** address the breaking change issue?

Alexander-Taran:
Option should be a well hidden one
I’m against options (-;

Unless it’s old-buggy-behavior-shim-do-not-use.ts

Bumping a major with a clear note
Can stay on the previous one

fkleuver:
So router 2.0 then?

Alexander-Taran:
Seems like it. Can also mark it as pre-release and ask ppl to test it out

huochunpeng:
I would like aurelia has a pre-release practice. Previously there were few bugs in new if-else feature in template-resources due to not tested enough before release, luckily they were addressed in time.

Aurelia is fast moving target. Pre-release can help hardening new features or dangerous fixes like this.

Issue #509, opened 2017-08-03
addRoute throws config error when list of routes provided


jfstephe:
**I'm submitting a bug report**

* **Library Version:**
1.1.0

* **Operating System:**
Linux Ubuntu (docker)

* **Node Version:**
v8.1.3

* **NPM Version:**
5.0.3

* **JSPM OR Webpack AND Version**
JSPM 0.16.53

* **Browser:**
Chrome

* **Language:**
Typescript

**Current behavior:**
Add a route using the following

```
    this.router.addRoute({
        route: ['routeA', 'routeB'],
        name: 'blah', moduleId: './pages/dashboard/Dashboard',
        nav: true, title: 'My title'
    });

```

Results in:

```
aurelia-router.js:1107 Uncaught Error: Invalid Route Config for "blah": You must specify a "route:" pattern.
    at validateRouteConfig (aurelia-router.js:1107)
    at AppRouter.addRoute (aurelia-router.js:896)
    at NavBar.execute.NavBar.addBlankDashboard (NavBar.ts:36)
    at CallScope.evaluate (aurelia-binding.js:1583)
    at Listener.callSource (aurelia-binding.js:5146)
    at eval (aurelia-binding.js:5170)
    at HTMLDocument.handleDelegatedEvent (aurelia-binding.js:3291)
```

It seems to be in the validation code but the addRoute function looks to also be expecting a string rather than an array.

**Expected/desired behavior:**
Not to throw an exception and work as the declarative config blocks work.

* **What is the expected behavior?**
addRoute to work like the config specified declaratively.

* **What is the motivation / use case for changing the behavior?**
Seems like it should always have worked like this.

Vheissu:
This has been raised before, here specifically: https://github.com/aurelia/router/issues/158 - even though it's an issue from 2015 the answer I believe still stands. @EisenbergEffect might want to weigh in on this, but that is intended behaviour that `addRoute` only allows you to specify a singular route.

EisenbergEffect:
Correct.

EisenbergEffect:
I'm not sure we can change it at this point. Seems like a breaking change for an scenario that could be handled with another existing API.

jfstephe:
Thanks for the other issue reference. The main problem I have with this is the Typescript definition of RouteConfig allows either an item or a list and yet the addRoute() call doesn't allow the list part. This is just asking for confusion/wasted time etc.
[...]

PR #576, opened 2018-02-19
fix(router): make addRoute split list of routes into multiple routeco…

fkleuver:
#509


fkleuver:
@davismj I just noticed the "merge options" PR creeped in with this one. You want me to split them up or need something else done in order for this to be ready for mergin?

Issue #419, opened 2016-10-06
Option to specify page title separator


fracz:
**I'm submitting a feature request**

The default separator of the page title [is set to pipe `|`](https://github.com/aurelia/router/blob/master/src/navigation-instruction.js#L235). However, the parameter cannot be currently overridden in any way. [The only place](https://github.com/aurelia/router/blob/master/src/navigation-instruction.js#L229) that uses this function does not use this argument.

The issue has already been discussed [in gitter](https://gitter.im/aurelia/Discuss?at=57f6b25a68f560d80ce32f7d) with a suggestion that the separator could be configured on child-router basis, too.

CasiOo:
Any suggestions where the separator should be defined?
Adding a `titleSeparator` to `RouterConfiguration` seems obvious to me, but the property probably also needs to be added to `NavModel`?

I would like the following implementation:
- Each child has the option to specify its own separator. If none is set, then use the parent separator. If none is set and at the root, then use the default value `' | '`.
- A globally set `titleSeparator` on the root RouterConfiguration that can be overriden in any child

Example configuration
```
export class App {
  configureRouter(config, router) {
    config.title = 'Aurelia';
    config.titleSeparator = ' - ';
    config.map([
      { route: ['', 'welcome'], name: 'welcome',      moduleId: 'welcome',      nav: true, title: 'Welcome' },
      { route: 'users',         name: 'users',        moduleId: 'users',        nav: true, title: 'Github Users' titleSeparator: ' > ' },
      { route: 'child-router',  name: 'child-router', moduleId: 'child-router', nav: true, title: 'Child Router' }
    ]);

    this.router = router;
  }
}
```

EisenbergEffect:
@CasiOo That sounds great. We've got a PR we are reviewing to make title generation pluggable. I think this would be a nice addition as well. Are you interested in working on it and submitting a PR?

CasiOo:
@EisenbergEffect  Yes I will find time to work on a PR :)
I'm not yet confident with child routes, as I have just kept using a single router, but I'll put some time into it.

nichgalea:
Any updates regarding this, please?

It's a good nice-to-have enhancement.

CasiOo:
I've started some initial work on this feature.
The interesting part is how the title should be built, you can see my updated logic here: https://github.com/CasiOo/router/blob/feat-title-separator/src/navigation-instruction.js#L236

I have some questions that need to be answered before I can move on with the implementation.
[...]

PR #583, opened 2018-03-20
feat(router): add configurable title separator

pndewit:
Partly solves https://github.com/aurelia/router/issues/419.
Title can now be set for each router by setting a property in the `configureRouter` function: `config.titleSeparator = ' - ';`.

This PR should solve most use cases, in the future we could for example look at making the separator configurable for each route by adding it to the NavModel.

### DIFF
--- a/src/navigation-instruction.js
+++ b/src/navigation-instruction.js
@@ -258,7 +258,7 @@ export class NavigationInstruction {
       let viewPortInstruction = this.viewPortInstructions[viewPortName];
 
       if (viewPortInstruction.childNavigationInstruction) {
-        let childTitle = viewPortInstruction.childNavigationInstruction._buildTitle(separator);
+        let childTitle = viewPortInstruction.childNavigationInstruction._buildTitle(titleSeparator);
         if (childTitle) {
           childTitles.push(childTitle);
         }
@@ -266,11 +266,11 @@ export class NavigationInstruction {
     }
 
     if (childTitles.length) {
-      title = childTitles.join(separator) + (title ? separator : '') + title;
+      title = childTitles.join(titleSeparator) + (title ? titleSeparator : '') + title;
     }
 
     if (this.router.title) {
-      title += (title ? separator : '') + this.router.transformTitle(this.router.title);
+      title += (title ? titleSeparator : '') + this.router.transformTitle(this.router.title);
     }
 
     return title;
--- a/src/navigation-instruction.js
+++ b/src/navigation-instruction.js
@@ -245,7 +245,7 @@ export class NavigationInstruction {
     }
   }
 
-  _buildTitle(separator: string = ' | '): string {
+  _buildTitle(): string {
     let title = '';
     let childTitles = [];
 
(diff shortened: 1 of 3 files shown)


src/navigation-instruction.js
```diff
@@ -245,7 +245,7 @@ export class NavigationInstruction {
     }
   }
 
-  _buildTitle(separator: string = ' | '): string {
+  _buildTitle(): string {
```
davismj: removing the argument is a potential breaking change

pndewit: fair enough, I will restore that one :)

pndewit:
The `_buildTitle` function should now be backwards compatible in a sense that if a separator was passed, it will take precedence over the router's configured titleSeparator property.

pndewit:
and fixed a boo-boo

pndewit:
Is this ready to be merged?

pndewit:
@davismj Can this be merged?

Alexander-Taran:
@pndewit don't rush it (-:

pndewit:
@Alexander-Taran 2 weeks now for an (IMHO) easy PR. The longer maintainers wait, the more demotivating it becomes for people creating the PR's to create more. Don't get me wrong here because I don't want to sound rude, but I just really want Aurelia to succeed and this is not really helping.

@davismj Is everything ok now?

davismj:
@pndewit I've been in your shoes and I understand your frustration. (see https://github.com/aurelia/templating-resources/pull/333).

That said, there are a couple things that make this not an easy PR. First, it deviates from the strategy I outlined here https://github.com/aurelia/router/issues/419#issuecomment-373607077. That doesn't mean it isn't good, but it means I need to look a bit closely. Second, there already exist extension points (here: https://github.com/aurelia/router/blob/master/src/router.js#L110) for modifying the router title. I need to take a minute to review whether or not this extension point is working, documented, and how it interacts with this particular feature.

I haven't forgotten your PR. I genuinely appreciate your hard work. It does seem like a good solution. I'm working hard to find the time to sit down and play with the PR and make sure its of the highest quality before I pull it in. I appreciate your patience.

pndewit:
@davismj fair enough and thanks for the update, next time you could give an update like this earlier as I was really under the impression the PR was just gathering dust. Keep up the great work and let me know if you need any assistance! 👍

pndewit:
@davismj I was just taking a closer look at
https://github.com/aurelia/router/blob/master/src/router.js#L110 and that function is being called for each child router title. The separator has no effect on this extension point.
This extension point is particularly being used for internationalising titles as can be seen in the Aurelia docs: http://aurelia.io/docs/routing/configuration#internationalizing-titles

Hope this helps.

pndewit:
@davismj any updates? You got around checking/validating my [comment](https://github.com/aurelia/router/pull/583#issuecomment-378829100)?
Let me know if you need help or info 😃

src/router-configuration.js
```diff
@@ -161,6 +162,7 @@ export class RouterConfiguration {
 
     if (this.title) {
       router.title = this.title;
+      router.titleSeparator = this.titleSeparator || ' | ';
```
davismj: If title is falsey, titleSeparator will not get exported to the router.

pndewit:
@davismj You are awesome, thx for the thorough research and explanation! Helps me with future PR's!

PR #586, opened 2018-04-01
typing(RouteConfig): ability to use class as route config

bigopon:
type definition to support typescript users


davismj:
These typings are not valid until https://github.com/aurelia/templating-router/pull/75 is implemented.

PR #592, opened 2018-04-28
Resolve redirect

### LINKED ISSUE #480: Configuring a route with "redirect" property causes BlueBird warnings

* **Library Version:**
1.0.1

* **Operating System:**
Windows 10

* **Browser:**
all

* **Language:**
all 

**Current behavior:**
specifying "redirect" property in a route configuration, causes BlueBird warnings about "Warning: a promise was rejected with a non-error: [object Object]"

This comes from the router method:
```
  function _buildNavigationPlan(instruction, forceLifecycleMinimum) {
    var prev = instruction.previousInstruction;
    var config = instruction.config;
    var plan = {};

    if ('redirect' in config) {
      var redirectLocation = _resolveUrl(config.redirect, getInstructionBaseUrl(instruction));
      if (instruction.queryString) {
        redirectLocation += '?' + instruction.queryString;
      }

      return Promise.reject(new Redirect(redirectLocation));
    }
```

I tired creating a gist given the link in the Issue Template, however i can't replicate it in a Gist - i'm assuming the bluebird promises library isn't included in gists?

Not to create a warning in the console everytime that specific route is redirected

Less warnings in the console

I understand that this is a BlueBird "quirk", however I'm posing the question as to whether Aurelia adjusts its use of BlueBird to stop these warnings? or if we need to go to BlueBird and propose a "don't warn me when a promise is rejected with a non-error" flag?

fkleuver:
Should fix #480

Use `Promise.resolve` instead of `Promise.reject` when handling a redirect.

`BuildNavigationPlanStep.run` will check if the returned value from `.then()` is a `Redirect`, and if so, and call `next.cancel`.

The tests still pass but it kind of feels like I'm missing something. It seems "too easy". Maybe `LoadRouteStep.run` also needs to be changed (loop through the values and if any of them is a `Redirect`, call `next.cancel`).

@davismj Do you know if it's possible for a redirect to end up in `LoadRouteStep` but not in `BuildNavigationPlanStep`?

### DIFF
--- a/doc/article/en-US/router-configuration.md
+++ b/doc/article/en-US/router-configuration.md
@@ -58,6 +58,8 @@ To use Aurelia's router, your component view must have a `<router-view></router-
   * Parameterized routes match against a string with a `:token` parameter (ie: 'users/:id/detail'). An object with the token parameter's name is set as property and passed as a parameter to the route view-model's `activate()` function.
   * A parameter can be made optional by appending a question mark `:token?` (ie: `users/:id?/detail` would match both `users/3/detail` and `users/detail`). When an optional parameter is missing from the url, the property passed to `activate()` is `undefined`.
   * Wildcard routes are used to match the "rest" of a path (ie: files/*path matches files/new/doc or files/temp). An object with the rest of the URL after the segment is set as the `path` property and passed as a parameter to `activate()` as well.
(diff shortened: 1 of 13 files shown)


davismj:
Thanks for the hard work Fred. I've never heard of a redirect occurring in the load route step.

PR #594, opened 2018-05-10
doc(router-configuration): add webpack configuration

greensponge:
There is currently no place in the router-configuration documentation that describes how to resolve moduleId when using webpack.

Closes https://github.com/aurelia/framework/issues/883


davismj:
@greensponge thanks so much!

PR #600, opened 2018-05-16
fix(router-configuration): throw early on invalid pipeline steps

doktordirk:
An error with Router.addPipelineStep leads to the usual 'key/value cannot be null or undefined. Are you trying to inject/register something that doesn\'t exist with DI?' at a later stage, making debugging difficult. 

I suggest to add the test to addPipelineStep which lets the error appear where it was created.


EisenbergEffect:
@davismj This seems reasonable to me. Can you review?

PR #601, opened 2018-06-01
Fix: Router ignored the root when creating hrefs

rluba:
Aurelia router generated absolute paths but did not append the root path (ie. `<base href="…">`). This is incorrect and caused all non-default click actions (eg. CMD+Click, Right-click-> "Open in new Tab/Browser", etc.) to navigate to the wrong URL. (And broke links for crawlers.)

Normal clicks only worked because they are intercepted by Aurelia and forwarded to `aurelia-browser-history`, which contains [a corresponding bug](https://github.com/aurelia/history-browser/pull/41) that treats all URLs as fragments instead of expecting the base path to be present in absolute paths.

This fix requires [the corresponding fix for `aurelia-browser-history`](https://github.com/aurelia/history-browser/pull/41) or the history will forward incorrect fragments to the router.

### DIFF
--- a/src/router.js
+++ b/src/router.js
@@ -224,7 +224,7 @@ export class Router {
     }
 
     this.isExplicitNavigation = true;
-    return this.history.navigate(_resolveUrl(fragment, this.baseUrl, this.history._hasPushState), options);
+    return this.history.navigate(_resolveUrl(fragment, this.baseUrl, this.history._hasPushState, this.history.root), options);
   }
 
   /**
@@ -279,8 +279,21 @@ export class Router {
     }
 
     let path = this._recognizer.generate(name, params);
-    let rootedPath = _createRootedPath(path, this.baseUrl, this.history._hasPushState, options.absolute);
-    return options.absolute ? `${this.history.getAbsoluteRoot()}${rootedPath}` : rootedPath;
+    let rootedPath = _createRootedPath(path, this.baseUrl, this.history._hasPushState, this.history.root);
+    if (options.absolute) {
+      // Hamfisted workaround because history does not expose the origin without the root
+      let origin = this.history.getAbsoluteRoot();
+      if (this.history._hasPushState) {
+        if (this.history.root) {
+          origin = origin.substring(0, origin.length - this.history.root.length);
+        } else {
+          origin = origin.substring(0, origin.length - 1);
+        }
+      }
+
+      rootedPath = origin + rootedPath;
+    }
+    return rootedPath;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


rluba:
**Don’t merge this PR yet.**
I’ve discovered that this solution doesn’t work in all cases and that there are more issues when root is set to something else than `/`. I’ll test it more thoroughly and keep you updated.

rluba:
I discovered that redirects via route configurations, eg., `{route: '', name: 'home', redirect: 'somepath'}`, are broken when using my PR. But while trying to fix this, **I discovered that these redirects are _generally broken_ on master** (as of 5fcb6f4a0a9c4e552f9894b517104a4b336ac760).

There are multiple issues:

1. Since 487c33d50fa938c9e725180877f6510ecae89db5: `TypeError: Cannot read property 'childRouter' of undefined` is thrown because [`determineWhatToLoad(…)` tries to interpret the `Redirect` plan as a regular plan](https://github.com/aurelia/router/blob/6912e0af451c29b14c8acaee0e2cc19f7cd8a42a/src/route-loading.js#L36-L38) .
2. Since aafa070dece5de190b83c4cc56791d53f04a0acd: [`_buildNavigationPlan(…)` calls `_createNavigationInstruction(…)` and then `generate(…)` on a router](https://github.com/aurelia/router/blob/fb09343978650eaf24020dab3299a9b70c71323b/src/navigation-plan.js#L30) _before it knows it’s `baseURL`_, causing it to pass incorrect URLs to `new Redirect(…)`. This probably cancels out when there’s no `root` set, but it prevents me from fixing the redirect when a `root` is set. 

Are these issues already known?

rluba:
1. is fixed in a separate PR (#603)
2. is fixed in 0737128

The test coverage for this module is pretty horrible. I’ll help bring it up in the future if we decide to stick with Aurelia.

davismj:
> The test coverage for this module is pretty horrible. I’ll help bring it up in the future if we decide to stick with Aurelia.

That is a generous assessment lol. Part of the reason Router dev has been slow recently is [I find myself spending more time fixing the tests](https://github.com/aurelia/router/search?q=test&type=Commits) than getting to actual issues, and its very time consuming.

Thanks for your hard work. These are on my radar.

mreiche:
Can someone please update this PR and merge it finally? This requirement is still unsatisfied.

bigopon:
@mreiche We are waiting for this PR at #632 to get merged, and then I'll incorporate @rluba changes in a new PR, or would be even better if @rluba could create PR based on refactored code.

mreiche:
#632 has been merged a while ago.

PR #602, opened 2018-06-02
Feature: cross-module devmode

rluba:
This PR adds a task `gulp dev [--target <module>]`. This supports cross-module development by allowing you to test changes of one module in the context of another module.

This solves [the problems discussed here](https://discourse.aurelia.io/t/how-do-you-setup-your-development-environment-for-core-modules/1271?u=rluba).

An example: Run `gulp dev --target ../my-project` to work on `aurelia-router` and test all changes in the context of `my-project` that’s installed in a sibling directory. Whenever you make any changes to any files in `aurelia-router`…
1. it is automatically recompiled and
2. the result is copied into the `aurelia-router` module within `my-project`.

If you use webpack in watch-mode on `my-project` (eg. in a second shell tab), you can immediately test any changes:

1. Make changes to `aurelia-router`
2. Hit save
3. Reload you browser
4. Test your changes and repeat

The current state of this PR is intended as a starting point for discussions and by no means a complete solution. In my opinion it’s already way better than the status quo, but it lacks a lot of robustness checks:

- find `node_modules` though npm configuration instead of expecting it at `<root>/node_modules`,
- same for jspm
- handle fs stream errors
- handle fs.stat errors properly
- handle glob errors properly

What do you think? If you like this solution, we can roll it out on all aurelia projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,9 +33,28 @@ To build the code, follow these steps.
   ```shell
   gulp build
   ```
-5. You will find the compiled code in the `dist` folder, available in three module formats: AMD, CommonJS and ES6.
+5. You will find the compiled code in the `dist` folder, available in various module formats: AMD, CommonJS, ES6 …
+
+  Use `gulp build --format <format1> [--format <format2> …]` to build only certain formats.
 
 6. See `gulpfile.js` for other tasks related to generating the docs and linting.
+
+### Continuious building and cross-module development
+
+  To build the dist files whenever you make any changes, run:
+
+  ```shell
+  gulp dev
+  ```
+
+  If you want to test your changes in the context of another module that depends on `aurelia-router`, run
+
+  ```shell
+  gulp dev --target <target_module_dir> [--format <format>]
+  ```
+
+  Whenever changes are compiled, they’ll automatically be copied to the `aurelia-router` module within the target module, e.g., `<target_module_dir>/node_modules/aurelia-router` or the jspm equivalent.
+
 
 ## Running The Tests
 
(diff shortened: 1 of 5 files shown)


bigopon:
Very nice, a feature request: while in dev mode, building project for all targets would be unnecessarily inefficient, it would be nice if it accepts argument to specify target:

```shell
gulp dev --target ../my-project --format es2015
```

rluba:
@bigopon Here you go. Also works for `gulp build` and you can specify multiple formats:

```shell
gulp build -f commonjs -f amd
```

bigopon:
@rluba I've been getting error with `gulp.watch` and it seems like we either need to upgrade to gulp version 4 or use `gulp-watch` instead. https://github.com/gulpjs/gulp/issues/1921 . And it doesn't copy on first run I think.

Have you experienced any issue watching the files ?

rluba:
What kind of error did you get? The one described in gulpjs/gulp#1921? I haven’t experienced any issues with `gulp.watch`so far - neither with this task nor any other.

> And it doesn't copy on first run I think.
Yes, that’s true – but trivial to fix. I can update the PR if you’re interested in merging it.

davismj:
@bigopon What do you think? I think this would be a good concept but I've not had a chance to play with it much. Is it worth the work or do we want to take notes and bring it forward to vNext?

bigopon:
For this PR specifically I've already "merged" (read steal and adjust) @rluba 's work in our TS rework. I also mentioned @rluba in that PR about it. I think we can close this PR. @rluba Can you have a look at the new tasks there related to cross module development ? https://github.com/aurelia/router/pull/615/files#diff-2624298877ec987a3e49f3a716b66329

rluba:
Sure, I’ll have a look. Thanks for "merging" my idea 😊

PR #603, opened 2018-06-05
Fix error introduced in 487c33d5 (when trying to fix #480)

### LINKED ISSUE #480: Configuring a route with "redirect" property causes BlueBird warnings

rluba:
This causes #480 to re-appear until a better handling for cancellations exists – but it’s still better than the errors caused by improper handling of redirects at the moment.

There’s no easy way to call `next.cancel` from all places that can create a redirect. I’d suggest changing redirects to add a magic value to an `Error` (eg. `error.redirect = new Redirect(…)`) and rejecting with that.

But this would require changes in several places that check for `output instanceof Error` to determine the kind of result.

### DIFF
--- a/src/navigation-plan.js
+++ b/src/navigation-plan.js
@@ -85,6 +85,9 @@ export function _buildNavigationPlan(instruction: NavigationInstruction, forceLi
               childInstruction,
               viewPortPlan.strategy === activationStrategy.invokeLifecycle)
               .then(childPlan => {
+                if (childPlan instanceof Redirect) {
+                  return Promise.reject(childPlan);
+                }
                 childInstruction.plan = childPlan;
               });
           });
--- a/src/route-loading.js
+++ b/src/route-loading.js
@@ -81,6 +81,9 @@ function loadRoute(routeLoader: RouteLoader, navigationInstruction: NavigationIn
 
           return _buildNavigationPlan(childInstruction)
             .then((childPlan) => {
+              if (childPlan instanceof Redirect) {
+                return Promise.reject(childPlan);
+              }
               childInstruction.plan = childPlan;
               viewPortInstruction.childNavigationInstruction = childInstruction;
 


fkleuver:
I had a vague feeling I was missing something, but I couldn't spot it. Of course it's child routers.

I'll try to beef up the test coverage in this area before having another crack at this. Thanks for investigating this further, your information helps a lot.

@jdanyow I agree with him, the original bug is much "better" than the new bug.

rluba:
With the changes in this PR it’s better than both the original state (because there are no warnings for redirects in app routers) and the current state (because there are no exceptions for redirects in child routers). So I think it’s worth merging instead of reverting 487c33d5.

fkleuver:
@jdanyow  @EisenbergEffect Please merge this. I broke production apps with this (sorry about that!). See https://github.com/aurelia/router/issues/480

Issue #378, opened 2016-07-27
Child Route activate() called before childRouter has refreshed baseUrl


tomtomau:
**I'm submitting a bug report**
- **Library Version:**
  1.0.0-rc1.0.5

**Current behavior:**

Originally in Gitter chat:

So I have a AppRouter (A) which has a Route (B) which is a ChildRouter under the likes of "/register", I want to dynamically throw the user around different routes in this child router. So I started by creating essentially an index route (C) for ChildRouter (B) (i.e. it handles the "") request. What I would ideally do, is, in the activate() method of the index route (C), return a promise which is a redirect to a different route (D) (which might be like "/step-1", so in sum it should be "/register/step-1"),as a route that belongs to ChildRouter (B). Catch is, if I try to router.navigateToRoute("route-d") sort of thing in the activate method of C, then it trys to redirect to the absolute app root ("/step-1" and not "/register/step-1"). If I call the same navigateToRoute method but later as part of click.delegate() call, it works perfectly. I cannot work out why it doesn't work in the activate method?

We established my approach was largely flawed anyway for the use case, but I can at least answer why I had this issue.

After debugging, I tracked it down to the activate() method of a child VM (C) of the child router (B) is being called before "refreshBaseUrl()" is called. Calling refreshBaseUrl() at the start of generate for example then breaks because the parent (AppRouter) is currently navigating and doesn't have a currentInstruction of it landing on the childRouter.

It might be a bit of a failing of my understanding of the router in that, I would have thought at the point that we were activate()'ing a one of the childRouter's child routes, that the AppRouter would have had a currentInstruction (ie, it landed on the childRouter).

Regardless, the behaviour as described below is not documented anywhere at present.

"During the activation of a VM belonging to a route within a ChildRouter, redirectToRoute on the child router will not generate the correct URL as the AppRouter has not resolved for the currentInstruction"

Happy to discuss further, would love to give an example but a bit time crunched at present but will aim to do so.

EisenbergEffect:
The only way to redirect during a navigation is to return a navigation comment (such as Redirect or RedirectToRouter) from the `canActivate` callback. Trying to do it later on or via a navigation api call will break things. Would you be interested in contributing by submitting some explanation to our documentation?

vuorinem:
@EisenbergEffect This doesn't seem to be working with child routes.

I created a gist: https://gist.run/?id=1bca105623d3c88400738985d008fc2e
[...]

Issue #605, opened 2018-06-20
[BUG] v1.6.x broke redirect in child router config


huochunpeng:
**I'm submitting a bug report**

* **Library Version:**
1.6.1

* **Please tell us about your environment:**
OSX 10.13

* **Node Version:**
8.11.2

* **NPM Version:**
6.1.0

* **JSPM OR Webpack AND Version**
NA

* **Browser:**
all

* **Language:**
ESNext

**Current behavior:**
With latest v1.6.1, when a route page `/child` has additional child router config, the redirect in config doesn't work properly.

```js
export class Child {
  configureRouter(config, router) {
    this.router = router;
    config.mapUnknownRoutes('./not-found.html');

    config.map([{
      route: 'test',
      title: 'test',
      name: 'test',
      nav: true,
      moduleId: './test.html'
    }, {
      route: '',
      // this redirects "/child" to "/child/test"
      // but in router 1.6.1, it instead redirects to "/test" (not-found)
      redirect: 'test'
    }]);
  }
}
```

**Expected/desired behavior:**
rjs-bundle is not up to date to router 1.6.x, so I built a demo app for the bug https://github.com/huochunpeng/router-bug

* **What is the expected behavior?**
Keep the old behaviour, redirect to the same level of route.

* **What is the motivation / use case for changing the behavior?**
:-( It broke my production app, I need to pin aurelia-router to 1.5.0 to bypass this regression.

bobum:
I can confirm this behavior in our app as well

EisenbergEffect:
@davismj Can you look into this. Appears to be a rather serious couple of issues in the latest router release.

fragsalat:
:+1:  Got this problem as well on main router after updating from 1.5.0 to 1.6.x

davismj:
On it.

balazsmeszegeto:
I was about to submit the very same issue

josundt:
Any progress on this?
Upgrading to 1.6.x is no option in my current project till this is OK

stefan505:
Guys, this issue is rather serious. Its completely broken my sub-routers + defaults strategy. 

As a long-time Aurelia user / supporter I am very concerned about the recent router problems.

huochunpeng:
@stefan505 for temporary fix, I use this in my package.json

```
"aurelia-router": "1.5.0",
"aurelia-templating-router": "1.3.2", // I need to do this for yarn. Didn't test on npm.
```

The regression is from #515, which can be reverted.

[...]

PR #610, opened 2018-07-16
Fix/child redirect

davismj:
router.baseUrl is used to generate urls for a route configured on any given router. The baseUrl is set by a router's current parent instruction. In light of this, it was originally designed to update the baseUrl when the parent instruction was fully processed and accepted, on the commitChanges step. However, this has always been a problem for child routers using redirect. They cannot use the generate method to generate a new URL because the baseUrl hasn't yet been updated.

This test demands a totally different approach. Whenever the router creates a new instruction the baseUrl should be updated. This is a well defined solution if we never enter a state where a baseUrl was updated to a value that was never finalized. This is never the case since a child router is--as far as I know--tied to the parent route. If the parent route fails to activate, then the child router is discarded. It will have an invalid value on it, but it will never be used for any purpose.

I don't anticipate any breaking changes. This should answer some issues of unexpected behavior with the router, including #378. If someone was relying on this behavior it may cause issues, but it was never a promised or documented behavior.

### DIFF
--- a/src/navigation-instruction.js
+++ b/src/navigation-instruction.js
@@ -198,7 +198,6 @@ export class NavigationInstruction {
 
     this.config.navModel.isActive = true;
 
-    router._refreshBaseUrl();
     router.refreshNavigation();
 
     let loads = [];
--- a/src/router.js
+++ b/src/router.js
@@ -454,8 +454,7 @@ export class Router {
 
   _refreshBaseUrl(): void {
     if (this.parent) {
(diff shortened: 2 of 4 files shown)


fkleuver:
Just a minor nit on the trailing/leading slash thing, other than that I don't see any issues either. I like this change, it's cleaner this way.

huochunpeng:
👍 passed both my test app and production app for fixing #605.

davismj:
@fkleuver what trailing / leading slash thing?

PR #613, opened 2018-08-07
fix(typings): Revert aliasing interface

bigopon:
Flow doesn't support type aliasing so the new interface was breaking `d.ts` generation. We haven't released any new typing with this type alias.

@EisenbergEffect @davismj

### DIFF
--- a/src/interfaces.js
+++ b/src/interfaces.js
@@ -212,8 +212,3 @@ interface PipelineResult {
   output: any;
   completed: boolean;
 }
-
-/**
-* The result of a router navigation.
-*/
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
@bigopon Looks like you might need to import `PipelineResult`.

bigopon:
@EisenbergEffect nice catch. `gulp build` ran fine so I thought that was it.

EisenbergEffect:
@davismj I went ahead and merged this since it's breaking builds and @bigopon needed it to move ahead with some of the work he's doing. Hope you don't mind!

PR #615, opened 2018-08-11
refactor(all): Convert to TS WIP

bigopon:
I think we should convert this repo to TS, as its current state is too error prawn

After this we should be able to iterate faster and safer.

@davismj @EisenbergEffect @fkleuver 

TODO: redo `gulp doc` task

@EisenbergEffect What gulp tasks do I have to do to convert this to rollup + TS ?


EisenbergEffect:
@bigopon I don't know anything about the roll-up work. Regarding the doc task, we don't need that in the repo anymore. It's not actually used to generate the docs. It's still useful for non-TS repos because I use that to validate that the output d.ts files are actually correct. For TS, they're guaranteed to be correct. However, the real API generation is still a problem.That's handled by the site-gnerator. It doesn't work for native TS libraries currently. So, we can't release this in TS without fixing that. FYI This is why there's no API documentation for Validation either.

bigopon:
@EisenbergEffect @davismj It's in good state now. All tests passed. I have to add this [line](https://github.com/aurelia/router/pull/615/files#diff-d8377f95ecc9f3984ff84c55ecb6a083R163) because the way code is compiled for test (to raw es2015, so this [line](https://github.com/aurelia/router/pull/615/files#diff-52fd7ff9c4a70144c285b45938c2af80R140) will fail without `lifecycleArgs` ). Beside that and some extra typings, all remain the same

package.json
```diff
@@ -62,6 +63,8 @@
     "aurelia-route-recognizer": "^1.2.0"
   },
   "devDependencies": {
+    "@types/jasmine": "^2.8.8",
+    "aurelia-templating": "^1.8.2",
```
EisenbergEffect: What's the need for adding the templating library here?

src/interfaces.ts
```diff
@@ -1,11 +1,35 @@
...
+
+/**@internal */
+declare module 'aurelia-dependency-injection' {
+  interface Container {
+    getChildRouter?: () => Router;
+  }
+}
+
+/**@internal */
+declare module 'aurelia-templating' {
```
EisenbergEffect: I understand why the above dependency was added. Bummer. This is a breaking change probably. We can live with it though.

src/interfaces.ts
```diff
@@ -106,93 +130,93 @@ interface RouteConfig {
...
 * An optional interface describing the determineActivationStrategy convention.
 */
-interface RoutableComponentDetermineActivationStrategy {
+export interface RoutableComponentDetermineActivationStrategy {
   /**
   * Implement this hook if you want to give hints to the router about the activation strategy, when reusing
   * a view model for different routes. Available values are 'replace' and 'invoke-lifecycle'.
   */
-  determineActivationStrategy: (params: any, routeConfig: RouteConfig, navigationInstruction: NavigationInstruction) => 'no-change' | 'invoke-lifecycle'| 'replace';
+  determineActivationStrategy(params: any, routeConfig: RouteConfig, navigationInstruction: NavigationInstruction): 'no-change' | 'invoke-lifecycle' | 'replace';
```
EisenbergEffect: Should the return type also use a `keyof`?

src/interfaces.ts
```diff
@@ -106,93 +130,93 @@ interface RouteConfig {
...
   */
-  configureRouter(config: RouterConfiguration, router: Router): Promise<void>|PromiseLike<void>|void;
+  configureRouter(config: RouterConfiguration, router: Router): Promise<void> | PromiseLike<void> | void;
 }
 
 /**
 * An optional interface describing the available activation strategies.
 */
-interface ActivationStrategy {
+export interface ActivationStrategy {
```
EisenbergEffect: Should we convert this to an enum?

package.json
bigopon: @EisenbergEffect It's for `import { CompositionContext } from 'aurelia-templating'` for better intellisense.

src/interfaces.ts
bigopon: Maybe not ... ? if we use it in this way, the code we deliver won't get affected ? Or we can just retype `CompositionContext` in this lib

src/interfaces.ts
bigopon: yes, I missed this. Should be `ActivationStrategy[keyof ActivationStrategy]`

src/interfaces.ts
bigopon: I was doing so, that's why you see `,` instead of `;` in its fields. I'm not sure about it, because it's already exported, and no1 uses this interface, only the `export const activationStrategy`

src/interfaces.ts
bigopon: Retype `CompositionContext` or a part of it only

bigopon:
@davismj @EisenbergEffect I removed the `aurelia-templating` dependency, and added a new interface [here]( https://github.com/aurelia/router/pull/615/commits/8e5d11bcc215cd5e3f68be2a068417a44eab4522#diff-a84f60e04d03dd963458bd8a12f9ac20R240) (It's compatible with `CompositionContext`)

For changes like this https://github.com/aurelia/router/pull/615/files#diff-52fd7ff9c4a70144c285b45938c2af80R9 , it's a semi breaking change: those classes aren't expected to be used outside of `router`, but if folks complain, we can change it to something like this:

```ts
  run(navigationInstruction: NavigationInstruction, next: Partial<Next>)
```

The `Partial` will make it compatible with `Function`

bigopon:
@rluba thanks for #602 , I stole the idea and added it in. Code doesn't look very nice but it somewhat does the job. If anyone has an idea to improve it, please send help 😄  to: https://github.com/aurelia/router/pull/615/files#diff-2624298877ec987a3e49f3a716b66329

@EisenbergEffect @davismj I also added new build / dev script in the README

src/interfaces.ts
fkleuver: This interface never made any sense to begin with because it describes an object containing activationStrategies, not an actual activationStrategy (which is just a plain string).
*This* would make sense: `export type ActivationStrategy = 'no-change' | 'invoke-lifecycle' | 'replace'`

What this interface *could* be useful as, is `HasActivationStrategy`. E.g. you could make a ViewModel implement it. But then still it doesn't make sense because it doesn't have a property `activationStrategy: 'no-change' | 'invoke-lifecycle' | 'replace'` (like it should).

Since the object (or enum, they're the same thing) are both capable of describing themselves, this interface is unquestionably redundant.

StrahilKazlachev: @fkleuver Actually the `interface ActivationStrategy` should describe a `const` similar to the `bindingMode` one. Though not sure where it is 😅

karma.conf.js
```diff
@@ -1,81 +1,50 @@
...
+      },
+      devtool: "cheap-module-eval-source-map",
+      module: {
+        rules: [
+          {
+            test: /\.ts$/,
+            loader: "ts-loader",
+            exclude: /node_modules/,
+            options: {
+              transpileOnly: true
```
fkleuver: I typically turn this on only when a code base is volatile and I just want tests to run even with typing errors. It also speeds up the compilation to have this on.
You might want to remove this or make it parameterized as a measure of robustness for this mature library. It's probably better to see typing errors during testing as well.

bigopon: Fixed, thanks !

bigopon:
@davismj Yes, but the classes exported (which have `Function` changed to `Next`) won't normally be used, which makes it a minor issue, so I think it's ok to process. But we have options to handle it.

src/activation.ts
```diff
@@ -55,14 +64,18 @@ function processDeactivatable(navigationInstruction: NavigationInstruction, call
   return iterate();
 }
 
-function findDeactivatable(plan, callbackName, list: Array<Object> = []): Array<Object> {
+function findDeactivatable(
+  plan: Record<string, ViewPortInstruction>,
+  callbackName: string,
+  list: Object[] = []
```
davismj: `IActivatableInfo[]`

src/activation.ts
```diff
@@ -132,36 +150,57 @@ function processActivatable(navigationInstruction: NavigationInstruction, callba
...
-      });
-    }
+  Object
+    .keys(plan)
+    .forEach((viewPortName) => {
+      let viewPortPlan = plan[viewPortName];
+      let viewPortInstruction = navigationInstruction.viewPortInstructions[viewPortName];
+      let viewModel = viewPortInstruction.component.viewModel;
+
+      if (
```
davismj: nitpick but this is not any easier to read and takes up more vertical space, which i see as a net negative

src/app-router.ts
```diff
@@ -14,6 +23,15 @@ const logger = LogManager.getLogger('app-router');
 export class AppRouter extends Router {
   static inject() { return [Container, History, PipelineProvider, EventAggregator]; }
 
+  /**@internal */
+  events: EventAggregator;
```
davismj: not internal, this is part of the public api

src/app-router.ts
```diff
@@ -1,10 +1,19 @@
...
-import {PipelineProvider} from './pipeline-provider';
-import {isNavigationCommand} from './navigation-commands';
-import {EventAggregator} from 'aurelia-event-aggregator';
+import { Container } from 'aurelia-dependency-injection';
+import { History, NavigationOptions } from 'aurelia-history';
+import { Router } from './router';
+import { PipelineProvider } from './pipeline-provider';
+import { isNavigationCommand } from './navigation-commands';
+import { EventAggregator } from 'aurelia-event-aggregator';
+import { NavigationInstruction } from '.';
```
davismj: ./navigation-instruction?

src/app-router.ts
```diff
@@ -54,16 +72,18 @@ export class AppRouter extends Router {
...
     if (!this.isActive) {
       let viewModel = this._findViewModel(viewPort);
       if ('configureRouter' in viewModel) {
         if (!this.isConfigured) {
           let resolveConfiguredPromise = this._resolveConfiguredPromise;
-          this._resolveConfiguredPromise = () => {};
-          return this.configure(config => viewModel.configureRouter(config, this))
+          this._resolveConfiguredPromise = () => { };
+          return this
+            // potential error here as callback function should return RouterConfiguration
```
davismj:
```
            .configure(config => {
              viewModel.configureRouter(config, this);
              return config;
            })
```

src/interfaces.ts
davismj: @StrahilKazlachev here https://github.com/aurelia/router/blob/master/src/navigation-plan.js#L6

@bigopon Sounds like we can remove this completely, right?

src/interfaces.ts
```diff
@@ -201,14 +217,62 @@ interface PipelineStep {
...
+
+  config?: RouteConfig;
+
+  moduleId?: string;
+
+  component?: RoutingComponent;
+
+  childRouter?: Router;
+
+  lifecycleArgs?: any[];// [Record<string, any>, RouteConfig, any]
```
davismj: [Record<string, string>, RouteConfig, NavigationInstruction]

src/interfaces.ts
```diff
@@ -201,14 +217,62 @@ interface PipelineStep {
...
+}
+
+export interface Viewport {
+  /**@internal */
+  container: Container;
+  swap(viewportInstruction: ViewPortInstruction): void;
+  process(viewportInstruction: ViewPortInstruction, waitToSwap?: boolean): void;
+}
+
+export interface ViewPortInstruction {
```
davismj: Having name be the only required part doesn't make sense to me. I don't think name is used anywhere at all. I believe strategy, config, moduleId, component, and lifecycleArgs are all required.

src/nav-model.ts
```diff
@@ -38,6 +41,9 @@ export class NavModel {
   */
   router: Router;
 
+  /**@internal */
```
davismj: I'm not sure if this should be internalized. I could see someone depending on it?

src/navigation-commands.ts
```diff
@@ -13,19 +27,28 @@ interface NavigationCommand {
...
+  shouldContinueProcessing: boolean;
+
+  private router: Router;
+
   /**
    * @param url The URL fragment to use as the navigation destination.
    * @param options The navigation options.
    */
-  constructor(url: string, options: any = {}) {
+  constructor(url: string, options: Partial<NavigationOptions> = {}) {
```
davismj: Why Partial?

src/navigation-instruction.ts
```diff
@@ -1,13 +1,18 @@
...
+import { ViewPortInstruction, RouteConfig, ActivationStrategy } from './interfaces';
+import { Router } from './router';
+import { activationStrategy } from './navigation-plan';
+
+export interface NavigationInstructionInit {
+  fragment?: string;
+  queryString?: string;
+  params?: Record<string, any>;
+  queryParams?: Record<string, any>;
+  config?: Partial<RouteConfig>;
```
davismj: Why partial? Why optional?

src/navigation-instruction.ts
```diff
@@ -1,13 +1,18 @@
...
+  parentInstruction?: NavigationInstruction;
+  previousInstruction?: NavigationInstruction;
+  router?: Router;
```
davismj: why optional

src/navigation-instruction.ts
```diff
@@ -1,13 +1,18 @@
...
-  parentInstruction: NavigationInstruction,
-  previousInstruction: NavigationInstruction,
-  router: Router,
-  options: Object
+import { ViewPortInstruction, RouteConfig, ActivationStrategy } from './interfaces';
+import { Router } from './router';
+import { activationStrategy } from './navigation-plan';
+
+export interface NavigationInstructionInit {
+  fragment?: string;
```
davismj: why optional?

src/navigation-instruction.ts
```diff
@@ -121,11 +133,11 @@ export class NavigationInstruction {
   /**
   * Adds a viewPort instruction.
   */
-  addViewPortInstruction(viewPortName: string, strategy: string, moduleId: string, component: any): any {
-    const config = Object.assign({}, this.lifecycleArgs[1], { currentViewPort: viewPortName });
-    let viewportInstruction = this.viewPortInstructions[viewPortName] = {
-      name: viewPortName,
-      strategy: strategy,
+  addViewPortInstruction(viewPortNameOrInstruction: string, strategy: string, moduleId: string, component: any): ViewPortInstruction {
```
davismj: why rename viewPortName. It is still typed as a string, and I don't know anywhere that passes in a full instruction, nor is there any handling for that.

src/navigation-instruction.ts
```diff
@@ -121,11 +133,11 @@ export class NavigationInstruction {
...
+    const config = Object.assign({}, this.lifecycleArgs[1], { currentViewPort: viewPortNameOrInstruction });
+    const viewportInstruction = this.viewPortInstructions[viewPortNameOrInstruction] = {
+      name: viewPortNameOrInstruction,
+      strategy: strategy as any,
```
davismj: if strategy argument is typed correctly this can be just strategy.

src/navigation-instruction.ts
```diff
@@ -204,7 +217,7 @@ export class NavigationInstruction {
     let delaySwaps = [];
 
     for (let viewPortName in this.viewPortInstructions) {
-      let viewPortInstruction = this.viewPortInstructions[viewPortName];
+      let viewPortInstruction: ViewPortInstruction = this.viewPortInstructions[viewPortName];
```
davismj: unnecessary

src/navigation-plan.ts
```diff
@@ -33,7 +41,7 @@ export function _buildNavigationPlan(instruction: NavigationInstruction, forceLi
         for (let key in newInstruction.params) {
           // If the param on the redirect points to another param, e.g. { route: first/:this, redirect: second/:this }
           let val = newInstruction.params[key];
-          if (typeof(val) === 'string' && val[0] === ':') {
+          if (typeof (val) === 'string' && val[0] === ':') {
```
davismj: nitpick but if you're going to add the space might as well drop the parens

src/pipeline-provider.ts
```diff
@@ -1,20 +1,29 @@
...
 
 
 class PipelineSlot {
+
+  /**@internal */
+  container: Container;
+  /**@internal */
+  slotName: string;
+  /**@internal */
+  slotAlias: string;
```
davismj: optional

src/pipeline-provider.ts
```diff
@@ -1,20 +1,29 @@
...
+
   steps = [];
 
-  constructor(container, name, alias) {
+  constructor(container: Container, name: string, alias: string) {
```
davismj: alias?: string

src/pipeline-provider.ts
```diff
@@ -29,7 +38,12 @@ class PipelineSlot {
 * Class responsible for creating the navigation pipeline.
 */
 export class PipelineProvider {
+
   static inject() { return [Container]; }
+  /**@internal */
+  container: Container;
+  /**@internal */
+  steps: (Function | PipelineSlot)[];
```
davismj: I don't think steps is ever a function.

src/pipeline-provider.ts
```diff
@@ -58,8 +72,9 @@ export class PipelineProvider {
     return pipeline;
   }
 
-  _findStep(name: string) {
-    return this.steps.find(x => x.slotName === name || x.slotAlias === name);
+  /**@internal */
+  _findStep(name: string): PipelineSlot {
+    return this.steps.find((x: PipelineSlot) => x.slotName === name || x.slotAlias === name) as PipelineSlot;
```
davismj: fix this.steps type and you won't need the PipelineSlot typings here

src/pipeline-provider.ts
```diff
@@ -68,8 +83,8 @@ export class PipelineProvider {
   addStep(name: string, step: PipelineStep): void {
     let found = this._findStep(name);
     if (found) {
-      if (!found.steps.includes(step)) { // prevent duplicates
-        found.steps.push(step);
+      if (!(found as PipelineSlot).steps.includes(step)) { // prevent duplicates
```
davismj: unneccessary typing

src/pipeline-provider.ts
```diff
@@ -68,8 +83,8 @@ export class PipelineProvider {
...
+        (found as PipelineSlot).steps.push(step);
```
davismj: unnecessary typing

src/route-loading.ts
```diff
@@ -0,0 +1,145 @@
...
+  }
+}
+
+export class LoadRouteStep {
+
+  static inject() { return [RouteLoader]; }
+  /**@internal */
+  routeLoader: RouteLoader;
+
+  constructor(routeLoader: RouteLoader) {
```
davismj: is there an advantage to @internal over 

`constructor(private routeLoader: RouteLoader)`?

src/route-loading.ts
```diff
@@ -0,0 +1,145 @@
...
+  let loadPromises = toLoad.map((current) => loadRoute(
+    routeLoader,
+    current.navigationInstruction,
+    current.viewPortPlan
+  ));
+
+  return Promise.all(loadPromises);
+}
+
+interface ILoadingPlan {
```
davismj: Maybe shift to interfaces

src/router-configuration.ts
```diff
@@ -16,11 +18,14 @@ export class RouterConfiguration {
...
     silent?: boolean;
   } = {};
-  pipelineSteps: Array<{name: string, step: Function|PipelineStep}> = [];
+  pipelineSteps: Array<{ name: string, step: Function | PipelineStep }> = [];
   title: string;
   titleSeparator: string;
-  unknownRouteConfig: string|RouteConfig|((instruction: NavigationInstruction) => string|RouteConfig|Promise<string|RouteConfig>);
-  viewPortDefaults: {[name: string]: {moduleId: string|void; [key: string]: any}};
+  unknownRouteConfig: string | RouteConfig | ((instruction: NavigationInstruction) => string | RouteConfig | Promise<string | RouteConfig>);
+  viewPortDefaults: { [name: string]: { moduleId: string | void;[key: string]: any } };
```
davismj: nitpick but you lost the space after the semicolon, maybe break this out onto multi lines

src/router-configuration.ts
```diff
@@ -110,7 +115,7 @@ export class RouterConfiguration {
    *  default, of the form { viewPortName: { moduleId } }.
    * @chainable
    */
-  useViewPortDefaults(viewPortConfig: {[name: string]: {moduleId: string; [key: string]: any}}): RouterConfiguration {
+  useViewPortDefaults(viewPortConfig: { [name: string]: { moduleId: string;[key: string]: any } }): RouterConfiguration {
```
davismj: nitpick but lost space after semicolon

src/router-configuration.ts
```diff
@@ -193,8 +198,9 @@ export class RouterConfiguration {
 
       let pipelineProvider = router.pipelineProvider;
       for (let i = 0, ii = pipelineSteps.length; i < ii; ++i) {
-        let {name, step} = pipelineSteps[i];
-        pipelineProvider.addStep(name, step);
+        let { name, step } = pipelineSteps[i];
+        // Potential error, addStep only accepts PipelineStep, but step here may be also a function
```
davismj: should never be a function, as far as I know.

src/router.ts
```diff
@@ -111,6 +143,23 @@ export class Router {
   */
   viewPortDefaults: any = {};
 
+  /**@internal */
+  catchAllHandler: (instruction: NavigationInstruction) => NavigationInstruction | Promise<NavigationInstruction>;
```
davismj: not internal

src/router.ts
```diff
@@ -111,6 +143,23 @@ export class Router {
...
+  /**@internal */
+  fallbackRoute: string;
```
davismj: not internal

src/interfaces.ts
```diff
@@ -201,14 +217,62 @@ interface PipelineStep {
...
+  childRouter?: Router;
+
+  lifecycleArgs?: any[];// [Record<string, any>, RouteConfig, any]
+
+  prevComponent?: RoutingComponent;
+
+  prevModuleId?: string;
+}
+
+export type NavigationResult = boolean | Promise<PipelineResult> | Promise<PipelineResult | boolean>;
```
davismj: doesn't `Promise<PipelineResult | boolean>` imply `Promise<PipelineResult>`?

src/router.ts
```diff
@@ -191,25 +240,27 @@ export class Router {
   *
   * @param callbackOrConfig The [[RouterConfiguration]] or a callback that takes a [[RouterConfiguration]].
   */
-  configure(callbackOrConfig: RouterConfiguration|((config: RouterConfiguration) => RouterConfiguration)): Promise<void> {
+  configure(callbackOrConfig: RouterConfiguration | ((config: RouterConfiguration) => RouterConfiguration)): Promise<void> {
     this.isConfigured = true;
 
-    let result = callbackOrConfig;
-    let config;
+    let result: RouterConfiguration = callbackOrConfig as RouterConfiguration;
```
davismj: It might be a config, or a function that returns void, config, promise<void>, promise<config>. Might be easiest to skip typing.

src/router.ts
```diff
@@ -278,6 +329,7 @@ export class Router {
       throw new Error(`A route with name '${name}' could not be found. Check that \`name: '${name}'\` was specified in the route's config.`);
     }
 
+    debugger;
```
davismj: rip

src/router.ts
```diff
@@ -289,7 +341,12 @@ export class Router {
   * @param config The route config.
   */
   createNavModel(config: RouteConfig): NavModel {
-    let navModel = new NavModel(this, 'href' in config ? config.href : config.route);
+    let navModel = new NavModel(
+      this,
+      'href' in config
+        ? config.href
+        // potential error when config.route is a string[] ?
```
davismj: looks like it to me too

src/router.ts
```diff
@@ -491,7 +556,9 @@ export class Router {
       let first = results[0];
       let instruction = new NavigationInstruction(Object.assign({}, instructionInit, {
         params: first.params,
-        queryParams: first.queryParams || results.queryParams,
+        queryParams: first.queryParams
+          // How is this valid
+          || (results as any).queryParams,
```
davismj: It's kind of a funny class. It doesn't inherit from Array but it adopts some of the Array methods right off the prototype so it looks and feels like an array but isn't.

Assuming this is typed correctly above, you shouldn't need the `as any`

https://github.com/aurelia/route-recognizer/blob/master/src/route-recognizer.js#L252

davismj: it is mistyped in route recognizer. do (as RecognizeResults) for now.

src/router.ts
```diff
@@ -505,7 +572,10 @@ export class Router {
     } else if (this.catchAllHandler) {
       let instruction = new NavigationInstruction(Object.assign({}, instructionInit, {
         params: { path: fragment },
-        queryParams: results ? results.queryParams : {},
+        queryParams: results
+          // how is this valid
+          ? (results as any).queryParams
```
davismj: (as RecognizeResults)

src/router.ts
```diff
@@ -518,7 +588,10 @@ export class Router {
 
         let instruction = new NavigationInstruction(Object.assign({}, instructionInit, {
           params: { path: fragment },
-          queryParams: results ? results.queryParams : {},
+          queryParams: results
+            // how is this valid
+            ? (results as any).queryParams
```
davismj: (as RecognizeResults)

src/router.ts
```diff
@@ -366,7 +427,8 @@ export class Router {
       }
 
       this.navigation.push(navModel);
-      this.navigation = this.navigation.sort((a, b) => a.order - b.order);
+      // this is a potential error / inconsistency between browsers
```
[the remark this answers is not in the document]
bigopon: I was unclear here. Only returning `0` will cause inconsistency behavior, per spec. From MDN:

> If compareFunction(a, b) returns 0, leave a and b unchanged with respect to each other, but sorted with respect to all different elements. Note: the ECMAscript standard does not guarantee this behaviour, and thus not all browsers (e.g. Mozilla versions dating back to at least 2003) respect this.

Because of that, if we have a list of route configs that all have `nav: true`, it will fall in this `return 0` behavior, thus may yield inconsistency

src/router.ts
bigopon: Maybe this ?

https://github.com/aurelia/router/commit/0cd5423a421c17d6bf0b10598ea451902ec5a590#diff-15ee858ae5d5aa9b3e2846cfd1c4c466R303

But it still feels a bit awkward to me, maybe we need to improve this a bit

src/router.ts
bigopon: lol

src/router-configuration.ts
bigopon: auto format of vscode x_x

src/route-loading.ts
bigopon: if we used `private`, the `d.ts` will still have it, while `@internal` strips it entirely from the generated result. It doesn't really matter much unless we have a lot of comments for `private` for better understanding from interfaces

src/pipeline-provider.ts
bigopon: Was because `found` could be either `PipelineSlot` or `Function` from existing type definition

src/pipeline-provider.ts
bigopon: yes, I missed this

src/navigation-instruction.ts
bigopon: Got the typing in before `this.viewPortInstructions`, Removing next commit

src/navigation-instruction.ts
bigopon: I was basing this on an attempt of overloading `addViewPortInstruction` and deprecate the current one for static router. Forgot to revert the name. for `strategy`, I wanted to but wasn't sure at first as it was a break change, but again, I think it's semi private and internal, probably fine to convert to keyof

src/app-router.ts
bigopon: nice

src/app-router.ts
bigopon: they was all rexported from index, I think that's why I got this automatically completed from the wrong module

src/router.ts
bigopon: @davismj not public either ? I don't see it in the current `aurelia-router.d.ts`

src/router.ts
bigopon: Same for this, it's not in the router in our current `aurelia-router.d.ts`

src/pipeline-provider.ts
bigopon: In the constructor of this class, we are adding some step constructors for later instantiation. like this [line](https://github.com/aurelia/router/pull/615/files/f41a488866ce19b1e3e2bc859262665818990519#diff-e0658779b23fb0d91b6c56ce52c62769R53)

src/pipeline-provider.ts
bigopon: Edit: It was necessary because of [this](https://github.com/aurelia/router/pull/615/files/f41a488866ce19b1e3e2bc859262665818990519#diff-e0658779b23fb0d91b6c56ce52c62769R53)

src/activation.ts
bigopon: I see that it's not easier to read, but the current state doesn't seem to help either, so I picked the 80 chars a line rule and go with it. We can have this line as an exception though, what do you think about a tslint file

src/navigation-commands.ts
bigopon: Here is what I need to work a bit more, `Redirect` is a commonly used class, and the existing typing for 2nd param is any. I need to make the new typing strong, while ensuring backward compat. I thought `Partial` would do the job but it seems I will need to do something else

src/nav-model.ts
bigopon: Mainly added this in typing for internal usage and it does not exist in our current distribution, I think we may not need to worry about it.

src/interfaces.ts
bigopon: I needed to make it partial to get it to work as easily as possible. Will improve the typing here

bigopon:
@davismj 

https://github.com/aurelia/router/pull/615/files#diff-52fd7ff9c4a70144c285b45938c2af80L44

The above block starting from that line seems to be wrong to me. It should be changed

from:

```js
let viewModel = infos[i];
let result = viewModel[callbackName](navigationInstruction);
return processPotential(result, inspect, next.cancel);
```

to

```js
let current= infos[i];
let result = current.viewModel[callbackName](navigationInstruction);
return processPotential(result, inspect, next.cancel);
```

Similar to this block

https://github.com/aurelia/router/pull/615/files#diff-52fd7ff9c4a70144c285b45938c2af80L121

src/route-loading.ts
bigopon: The only function that uses this interface is local to this module so I thought it may as well sit in the same place

src/interfaces.ts
bigopon: Yes, probably. I added a dedicated interface
https://github.com/aurelia/router/blob/a0135eba04cdf222795fff2c58cedd6a1987969e/src/interfaces.ts#L210

src/router-configuration.ts
bigopon: It's true that it cant be a function here, the existing typing is inaccurate.

src/interfaces.ts
davismj: @bigopon Should this be removed?

src/nav-model.ts
davismj: Order is part of the public api. See here for example: https://aurelia.io/docs/routing/configuration#basic-configuration

src/pipeline-provider.ts
```diff
@@ -110,7 +126,8 @@ export class PipelineProvider {
     this._clearSteps('postRender');
   }
 
-  _createPipelineSlot(name, alias) {
+  /**@internal */
+  _createPipelineSlot(name: string, alias?: string) {
```
davismj: This always returns a PipelineSlot

src/pipeline-provider.ts
davismj: They are all PipelineSteps. The placeholders are also PipelineSteps.

src/pipeline-provider.ts
davismj: I never was able to see what you were linking to. Do you know what its necessary?

src/router.ts
davismj: @bigopon

src/pipeline-provider.ts
bigopon: Somehow the old commits are gone. It's necessary because the way it works with container. If we gives it a Function, it will ask the container to create an instance of it, otherwise, as object, we assume it's an object with `PipelineSlot` shape

bigopon:
@davismj I've fixed most of the issues except this
```ts
steps: (Function | PipelineSlot)[];
```
Right in the constructor of it, we see how `Function` is a valid type for it.

bigopon:
@davismj I've merged all changes in master into my PR branch. It's good for final review / merge.

bigopon:
@davismj I've merged all changes in master into my PR branch. It's good for final review / merge.
cc @EisenbergEffect @fkleuver

bigopon:
The interface of `component` in view port has become strict, this may, or may not be an issue. If folks complain, we can turn  it into an index-able interface to avoid typing issue.

src/pipeline-provider.ts
davismj: For whatever reason, I was thinking `ActivateNextStep` was an actual step and not the class constructor.

src/router.ts
davismj: We were having problems generating the router.d.ts, which probably explains this.

src/router.ts
bigopon: I'd vote for it to be like it is atm. We can always change it to public if necessary.

src/router.ts
bigopon: I'd vote for it to be like it is atm. We can always change it to public if necessary.

PR #616, opened 2018-08-15
fix: Prevent running CanDeactivatePreviousStep twice in same navigation

### LINKED ISSUE #45: canDeactivate() is triggered twice
canDeactivate() is triggered twice when you navigate to another route where canActivate returns  new Redirect('...').

Is there any workaround?

davismj:
To handle this, I've added a new parameter to the router called `couldDeactivate` which is set to true whenever the CanDeactivatePreviousStep completes and is reset on every navigation. If a redirect is returned from the CanActivateNextStep, then this value will be true and it is used to build a special pipeline that skips the CanDeactivatePreviousStep on the next pipeline run.

fixes #45

### DIFF
--- a/src/activation.js
+++ b/src/activation.js
@@ -25,7 +25,7 @@ export class ActivateNextStep {
   }
 }
 
-function processDeactivatable(navigationInstruction: NavigationInstruction, callbackName: string, next: Funcion, ignoreResult: boolean) {
+function processDeactivatable(navigationInstruction: NavigationInstruction, callbackName: string, next: Function, ignoreResult: boolean) {
   const plan = navigationInstruction.plan;
   let infos = findDeactivatable(plan, callbackName);
   let i = infos.length; //query from inside out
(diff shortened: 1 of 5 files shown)


src/activation.js
```diff
@@ -49,6 +49,8 @@ function processDeactivatable(navigationInstruction: NavigationInstruction, call
       }
     }
 
+    navigationInstruction.router.couldDeactivate = true;
```
bigopon: This will treat `canDeactivate` and `deactivate` the same way. Does that have any implications ?

davismj: No. I originally had a check for `canDeactivate` but it served no purpose and because of the naming of the variable `ignoreResult` I found that the coded was less readable. 

```
if (!ignoreResult) {
  navigationInstruction.router.couldDeactivate = true;
}
```

That's just gunna cause someone to waste 30 minutes figuring out what `ignoreResult` is and why it has anything to do with `couldDeactivate`.

PR #624, opened 2018-10-22
WIP feat(router): ability to use classe/module as view model

bigopon:
Changes:
- For build plan step: previously we determine activation strategy by comparing `moduleId` of previous instruction and next view port config of the same view port name. This was simple when it is only either `string` or `null`. And 2 different view model factories could actually point to the same view model class, it needs to be resolved in advance. To do this without destroying readability, I converted the `_buildNavigationPlan` to async function. This will lead to slightly bigger build, but we can start distributing `es2018` so folks can use native `async/await`. And I took the opportunity to convert a lot of promises to `async/await` for better overall readability too.
- For `NavigationInstruction.prototype.addViewPortInstruction`, overload it to accept a view port name and a partial view port config instead of 4 primitive paramters. This makes it easier to pass view model around and keep the API clean. (Should we also deprecate/remove the old API?)
- **LoadRouteStep**: With the addition of `viewModel` property on the `config`, `templating-router` Router needs to be updated to account for this. I will add a link to specific lines that address this in near future
- To be updated...

TODO:

- [ ] convert templating-router to TS and adopt the changes here
- [ ] unit test
- [ ] integration test

@davismj @EisenbergEffect


src/interfaces.ts
```diff
@@ -16,14 +16,16 @@ declare module 'aurelia-dependency-injection' {
...
+
 /**
 * A configuration object that describes a route.
 */
 export interface RouteConfig {
   /**
   * The route pattern to match against incoming URL fragments, or an array of patterns.
   */
-  route: string | string[];
+  route?: string | string[];
```
fkleuver: Why is this optional now?

bigopon: In view port route config, they are not required, I realized this while dealing with viewports
```js
{ route: 'a', viewports: { a: {}, b: {} }
```

src/app-router.ts
```diff
@@ -140,61 +140,59 @@ export class AppRouter extends Router {
...
+      restorePreviousLocation(this);
+      return this._dequeueInstruction(instructionCount + 1);
+    }
+    else if (instructionCount > this.maxInstructionCount) {
+      throw new Error('Maximum navigation attempts exceeded. Giving up.');
+    }
+    let pipeline = this.pipelineProvider.createPipeline(!this.couldDeactivate);
+    let result: PipelineResult;
+    try {
+      const $result = await pipeline.run(instruction);
```
fkleuver: Nice!

src/interfaces.ts
```diff
@@ -271,6 +283,7 @@ export interface ViewPortPlan {
 
   prevComponent?: ViewPortComponent;
   prevModuleId?: string;
+  prevViewModel?: string | Function | null;
```
fkleuver: In which way is a `viewModel` of type `string` different from `moduleId`?

bigopon: This could be changed. The interface is just a merge of `moduleId` and `viewModel`, as I was not sure how to pass things around nicely. Mostly to do something like this:
```js
if (typeof prevViewModel === 'string' || prevViewModel === null) {
  // it was moduleId
} else {
  // it was viewModel
}
```

src/navigation-instruction.ts
```diff
@@ -251,9 +266,8 @@ export class NavigationInstruction {
       .all(loads)
       .then(() => {
         delaySwaps.forEach(x => x.viewPort.swap(x.viewPortInstruction));
-        return null;
-      })
-      .then(() => prune(this));
+        prune(this);
```
fkleuver: You're skipping a tick here, is that intentional?

src/interfaces.ts
bigopon: null could also come from `viewModel` factory:
```js
viewModel: () => null
```
But I think we can treat it like `moduleId: null`

src/navigation-instruction.ts
bigopon: I think this was an accident converting this fn back and forth between promise and async/await. Thanks!

src/navigation-plan.ts
```diff
@@ -26,71 +29,78 @@ export class BuildNavigationPlanStep {
...
+    let redirectLocation = router.generate(newInstruction.config.name, params, instruction.options);
+    if (instruction.queryString) {
+      redirectLocation += '?' + instruction.queryString;
+    }
+    return new Redirect(redirectLocation);
   }
 
   const prev = instruction.previousInstruction;
-  const plan: Record<string, ViewPortPlan> = {};
+  const viewPortPlans: Record<string, ViewPortPlan> = {};
```
fkleuver: Nice :) these sorts of language details make the code much easier to follow

src/route-loading.ts
```diff
@@ -76,70 +100,77 @@ function determineWhatToLoad(
...
-                return loadNewRoute(routeLoader, childInstruction);
-              });
-          });
-      }
+  let config = viewPortPlan.config;
+  let moduleId: string | null = null;
+  let viewModel: ViewModelSpecifier | null = null;
+  if (config) {
+    if ('moduleId' in config) {
+      moduleId = config.moduleId;
```
fkleuver: Small detail here. This will take any moduleId, even if it's `null`/`undefined`/`''`, and overwrite the existing `null` value with it. Are empty string and undefined accounted for as potential values now?

test/activation.spec.ts
```diff
@@ -13,23 +13,26 @@ import {
 } from '../src';
 import { ValueOf, createPipelineState, MockPipelineState, MockInstruction } from './shared';
 
-describe('activation', () => {
-  describe('CanDeactivatePreviousStep', () => {
+describe('activation', function activation() {
```
fkleuver: just curious why did you name this function?

bigopon: In vscode outline view, we have nice view with named function:
image
(Each nested fn is a `describe`)
It aids navigation when the test file becomes too lengthy. I think we should do this for vNext too.

src/route-loading.ts
bigopon: It only comes when user defines it. So I'm just doing simple garbage in, garbage out here. In order to keep backward compat, `moduleId` > `viewModel`

fkleuver:
For readability/maintainability this is nothing but improvement, no question there. Some of this should have happened a long time ago, but it's good to see it happening now :)
I have a vague gut feeling about the switch to async/await that timings are going to be affected. It makes it feel like a somewhat risky PR. But the long term benefit can be really big. Looking forward to those integration tests 😆

test/activation.spec.ts
fkleuver: ooh really now! I had no idea. That's awesome

src/navigation-instruction.ts
bigopon: I was confused by this, and did some check, by playing with console. and also trying to read the spec and asking on StackOverflow. From what I've seen so far, chaining seems to be equivalent to synchronous call without any intermediate step.

fkleuver: Each `.then()` puts the code that follows in the next tick. So there's definitely a difference between the two.
This would surface when there are other promises going on in parallel (which there probably are). Now prune would come one tick earlier:

```ts
function prune(id) {
	console.log('prune', id);
}
function doSomething(id) {
	console.log('doSomething', id);
}
function pruneNextTick() {
	Promise.resolve()
		.then(() => {
			doSomething(1);
		})
		.then(() => {
			prune(2);
		})
	Promise.resolve()
		.then(() => {
			doSomething(3);
		})
}
function pruneSameTick() {
	Promise.resolve()
		.then(() => {
			doSomething(1);
			prune(2);
		})
	Promise.resolve()
		.then(() => {
			doSomething(3);
		})
}
```

`pruneNextTick() // simulates previous code`

```shell
doSomething 1
doSomething 3
prune 2
```

`pruneSameTick() // simulates current code`

```shell
doSomething 1
prune 2
doSomething 3
```

Whether it matters / breaks something is another question. Perhaps it doesn't matter. But there's a difference for sure

bigopon:
```js
Promise.resolve()
		.then(() => {
			doSomething(1);
		})
		.then(() => {
			prune(2);
		})
	Promise.resolve()
		.then(() => {
			doSomething(3);
		})
```
Isn't the the log the way it is because the way promise callback are enqueued, not because of the tick behavior? I'd expect at it to log exactly as your example showed.

src/navigation-plan.ts
```diff
@@ -101,43 +111,42 @@ export function _buildNavigationPlan(
...
-                if (childPlan instanceof Redirect) {
-                  return Promise.reject(childPlan);
-                }
-                childInstruction.plan = childPlan;
-                // for bluebird ?
-                return null;
-              });
+            );
+            if (childPlanOrRedirect instanceof Redirect) {
+              return Promise.reject(childPlanOrRedirect);
```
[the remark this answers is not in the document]
bigopon: They behave in the same way, so it's matter of preference imo

src/navigation-instruction.ts
fkleuver: You can write results to an array, result is the same. Regardless, whoever told you that this:
```ts
.then(() => a())
.then(() => b())
```
it the same as this:
```ts
.then(() => a(); b();)
```

Is wrong. If you asked on SO and they gave you this answer, tell me the link to the question and I'll tell them how wrong they are.

fkleuver: I just realized I slightly misread your message. I thought you implied there was different behavior when logging in the console.
I'm not sure what it matters whether it's about the way promises are queued vs the tick behavior. Aren't those 2 things the same fabric anyway?
Each `.then` is placed on the micro task queue after the previous code is done executing. If that previous code is synchronous, then it happens the next tick. Even if you put 1000 `Promise.resolve().then()` calls right below eachother (not chained), all 1000 will happen the next tick. But if you chain only one, the second `then()` happens the tick after the previous `then()`, so 2 ticks later..

src/interfaces.ts
davismj: moduleId is optional, route is not optional

bigopon: Yes, will need to change this. Probably because the way viewport typings is reused

bigopon:
@davismj I've rebased and updated dependency route recognizer. It's working good

tmueller:
Came here from https://discourse.aurelia.io/t/discussion-bundler-friendly-aurelia/959/17. Any news on this?

bigopon:
@tmueller This work has been merged. It's now waiting for some checks and reviews, and I'm working with @davismj here https://github.com/aurelia/templating-router/pull/84 to complete this feature.

bigopon:
Closing this as it has been merged in another PR.
Thanks to @tmueller for reminding

tmueller:
Thanks @bigopon for the fast response :)

PR #635, opened 2019-05-03
fix: fix regression on async configureRouter

huochunpeng:


huochunpeng:
@fkleuver what's meaning of the ci result? Do I need to push up updated dist/ files?

huochunpeng:
Ok, there is type error in build. I will revise.

huochunpeng:
There is TS typing hole out of my knowledge. @fkleuver pls fix this bug :-)

Issue #639, opened 2019-05-27
Wrong queryString when building navigation plan in aurelia-router


MartyBoi:
**I'm submitting a bug report**

* **Library Version:**
1.7.1

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
10.15.1

* **NPM Version:**
6.4.1

* **JSPM OR Webpack AND Version**
webpack 4.29.6

* **Browser:**
all

* **Language:**
TypeScript

**Current behavior:**
Not redirected with correct query parameter. We are using _aurelia-open-id-connect_ and when redirected back from our identity provider the query parameters are removed when creating the redirectPlan in _aurelia-router_,  _navigation-plan.ts_.
In my example you can see that the path we're navigating to is "/example?test=123". But in the end the query string is removed because the wrong instruction is used when getting query params.

![image](https://user-images.githubusercontent.com/50705141/58402414-89e1c080-8060-11e9-8adb-79b08a41ae87.png)

* **What is the expected behavior?**
Instead of getting queryString from _instruction_ like this:
```
if (instruction.queryString) {
    redirectLocation += '?' + instruction.queryString;
}
```
I think you should get it from _redirectInstruction_ like this:
```
if (redirectInstruction.queryString) {
    redirectLocation += '?' + newInstruction.queryString;
}
```

![image](https://user-images.githubusercontent.com/50705141/58402355-66b71100-8060-11e9-9c2e-c88ae6616fff.png)

bigopon:
@MartyBoi thanks for filing this issue. Any chance you could help with a minimal repro? Maybe just even pseudo code

MartyBoi:
@bigopon of course! Is this ok?
```
redirected from external site, to e.g "https://exampleurl.com/routePage?queryTest=value";
router captures with query string included;
router redirects to "routePage" with no query string, e.g "https://exampleurl.com/routePage";
```
Expected behavior:
```
redirected from external site, to e.g "https://exampleurl.com/routePage?queryTest=value";
router captures with query string included;
router redirects to "routePage" with query string included, e.g "https://exampleurl.com/routePage?queryTest=value";
```

bigopon:
@MartyBoi I tried to reproduce your issue with this test https://github.com/aurelia/router/pull/637/commits/24394dd6afac8c5b62ac6b1e81f4988445e8d796#diff-c70fb785178d11245e55f0a186031839R14

[...]

PR #643, opened 2019-07-01
fix(navigation-plan): add query string from redirect instruction

MartyBoi:
See issue: [https://github.com/aurelia/router/issues/639](url)

### DIFF
--- a/src/navigation-plan.ts
+++ b/src/navigation-plan.ts
@@ -81,6 +81,14 @@ export const buildRedirectPlan = (instruction: NavigationInstruction) => {
       let queryString = instruction.queryString;
       if (queryString) {
         redirectLocation += '?' + queryString;
+      }      
+      let redirectQueryString = redirectInstruction.queryString;
+      if (redirectQueryString) {
+        if (queryString) {
+          redirectLocation += '&' + redirectQueryString;
+        } else {
+          redirectLocation += '?' + redirectQueryString;
+        }
       }
 
       return Promise.resolve(new Redirect(redirectLocation));


bigopon:
@MartyBoi It doesn't seem we can simply append the redirect location. There're cases where things would override each other, and there is an issue with redirect with dynamic part in the URL too.

Would need @davismj to chime in

rmja:
This seems like a reasonable and non breaking change. Could it be merged?

GuodongFan:
test

rmja:
I would really appreciate if a maintainer would look at this PR. It seems really simple and I have used it for a long time now in a big app without any issues.
ping @bigopon

bigopon:
@rmja there' concerns stated in my comments above https://github.com/aurelia/router/pull/643#issuecomment-507198131
I don't think we can simply merge this yet. And for the very least, we should have a failing test, can you help with that? I can then help proceed this forward faster.

rmja:
@bigopon Thanks for getting back. There is a very simple example in the issue https://github.com/aurelia/router/issues/667.
Basically, if a route is configured like `{ route: "does-not-work", redirect: "/dashboard?autoselect=true"; }`, then the parameter `autoselect` is not passed on as a parameter to `dashboard`.

rmja:
How do we proceed with this?

rmja:
@bigopon ping..

bigopon:
Thanks @rmja , it's been a bit difficult to proceed with this for I haven't had time to focus on v1 repos (there been a few queued items), and this repo specifically needs special setup work refactoring to make all future work fast/easy. Will keep this in mind and act as soon as I can.

Issue #648, opened 2019-08-15
navigateToRoute return signature change - no longer a Promise?


matthewcorven:
- Overview of the Issue: router navigateToRoute() previously has a return signature type of `Promise<PipelineResult | boolean>`, [then](https://github.com/aurelia/router/commit/6b2232d8e7dd0420f50ce4a008b1a4b34264b84a) `NavigationResult` and [most recently](https://github.com/aurelia/router/commit/908ec620c136f1a4a3da5928ed751f5ee599c264) `boolean` (version 1.7.1).
- Motivation or Use Case: 
   - Upgrading from earlier versions to latest results in build failure in projects enforcing tslint rule "await-promise".
   - Code relying upon navigateToRoute as an awaitable function require refactoring
- Library Name and Version(s): aurelia-router, router.ts version 1.7.1 (current)
- Browsers and Operating System: N/A (build issue)
- Reproduce the Error: Build an application with lines below using aurelia-router version 1.7.1 and tslint "await-promise" rule turned on
- Related Issues: None found
- Suggest a Fix: Revert return signature change, or explain reason for change to return type `boolean`.

https://github.com/aurelia/router/blob/893b768f01aea842ee57db4222e66aa572f24404/dist/aurelia-router.d.ts#L300

EisenbergEffect:
Ping @davismj

bigopon:
If i understand correctly, returning a `Promise` is incorrect, since it's not what the history navigate returns

davismj:
The typing is in fact wrong. After quickly tracing through, I want to say it should be `boolean | Promise<PipelineResult>` but it might be `boolean | Promise<boolean | PipelineResult>`.

matthewcorven:
Thanks for the initial investigation @davismj

matthewcorven:
@davismj  @EisenbergEffect I'd be happy to submit a PR for this.  I'm not sure where the CLA is that I need to sign.  The CONTRIBUTING.md in this project sends me to:

http://goo.gl/forms/dI8QDDSyKR

Which says "Sorry, the file you have requested does not exist.":

EisenbergEffect:
Please feel free to submit a PR :) When you submit the PR, Github will check our CLA sig list and prompt you with the proper way to sign. Sorry about the bad link. I thought I had that updated everywhere after we switched over to CLA Assistant. I'll get that fixed soon.

Calabonga:
Hi, and again...

Aurelia version:
```
> au -v
[...]

PR #650, opened 2019-08-17
Change return signature of router navigate() and navigateToRoute() from boolean to NavigationResult

matthewcorven:
Fix for #648 to correct typings.  Calllers of navigate()/navigateToRoute() could await the routing to complete, but didn't know it due to typings.
@davismj


matthewcorven:
@davismj and perhaps type `NavigationResult` would be more clearly named a `AwaitableNavigationResult`?

test/router.spec.ts
```diff
@@ -230,6 +232,40 @@ describe('the router', () => {
           done();
         });
     });
+
+    it('should return to caller immediately if not awaited', (done) => {
```
davismj: well, these tests don't really test anything about aurelia. only how async works. they will never fail.

src/router.ts
```diff
@@ -310,7 +310,7 @@ export class Router {
    * @param params The route parameters to be used when populating the route pattern.
    * @param options The navigation options.
    */
-  navigateToRoute(route: string, params?: any, options?: NavigationOptions): boolean {
+  navigateToRoute(route: string, params?: any, options?: NavigationOptions): NavigationResult {
```
davismj: Strangely in my local version, this is what I already had. I wonder when and why they got changed.

test/router.spec.ts
matthewcorven: Good call, removed

davismj:
@matthewcorven thx for the PR, definitely helps me see what's up.

I think the fix needs to be a bit more robust here. It starts in history / browser where navigate has a boolean return type (https://github.com/aurelia/history/blob/master/src/index.js#L56, https://github.com/aurelia/history-browser/blob/master/src/browser-history.ts#L54), which is only accepted because of other insufficient typings around history options (https://github.com/aurelia/history-browser/blob/master/src/browser-history.ts#L111, https://github.com/aurelia/history-browser/blob/master/src/browser-history.ts#L297, https://github.com/aurelia/history-browser/blob/master/src/browser-history.ts#L54).

Additionally, I don't like the `NavigationResult` typing. Consuming one requires testing for a `Promise` first: `const result: boolean = await Promise.resolve(this.router.navigate('foo');`. This implementation detail should not be the burden of an end developer, who should simply be able to write `const result: boolean = await this.router.navigate('foo');`.

I'll go ahead and fix this in all three libraries by:
- Adding a typing for `NavigationResult` in aurelia-history
- Adding a typing for `HistoryOptions` in aurelia-history-browser
- Enforcing `Promise<NavigationResult>` return types on all navigation methods in all libraries

This will be a breaking change. Currently, the `NavigationResult` type includes `PipelineResult | boolean`, which is a difficult type to work with. I will unify them as `NavigationResult = { completed: boolean }`, and transform all flat `boolean` returns into a `NavigationResult`. This means, moving forward, any navigation results, no matter where they come from, can be consumed like this:

`const { completed } = await router.navigate('foo');`

matthewcorven:
Thanks @davismj !

PR #654, opened 2019-09-05
fix(app-router): include error stack trace in logger

baerrach:
closes aurelia/framework#926


baerrach:
I'm hacking this locally and am preferring this option instead

```
logger.error(result.output);
```

It doesn't double print the error message string and I still get the stack trace.

baerrach:
Also, the output may or may not give object expansion toggles in your browser console. It appears to be browser specific.

This may only be a problem if you attach custom properties to your Errors and expect them to be inspectable in the console.

A workaround is to write your own implementation of the logger and use `console.dir()` if there's an error.

